PR #20, opened 2021-11-04
connection quality api

davidliu:


davidzhao (approved):
looks good! Question: how do we surface this event to client apps? Wouldn't we need listeners?

davidliu:
> looks good! Question: how do we surface this event to client apps? Wouldn't we need listeners?

I'm thinking we'll need to move to an event bus similar to the js sdk, given the number of events and locations these changes are coming from.

PR #22, opened 2021-11-08
android higher level track apis

davidliu:


hiroshihorie (approved):
✅

davidzhao:
what ends up calling `setTrackEnabled` ? Could you update the README docs as well?

davidliu:
Agh, forgot to include the direct calls for the camera/mic. good catch. will update.

davidzhao:
it'd be great to update the sample app to use them as well.

derek-m-ir:
No longer able to mute/unmute LocalTrackPublication.  The mute property is read-only

PR #23, opened 2021-11-11
android automatic track management

davidliu:


livekit-android-sdk/src/main/java/io/livekit/android/room/track/Track.kt
```diff
@@ -1,5 +1,8 @@
 package io.livekit.android.room.track
 
+import io.livekit.android.events.BroadcastEventBus
+import io.livekit.android.events.EventListenable
```
boks1971, line 4: Is this import used?

livekit-android-sdk/src/main/java/io/livekit/android/room/track/video/VideoSinkVisibility.kt
```diff
@@ -0,0 +1,136 @@
...
+        val height = lastCoordinates?.size?.height ?: 0
+        return Track.Dimensions(width, height)
+    }
+
+    fun onGloballyPositioned(layoutCoordinates: LayoutCoordinates) {
+        val lastVisible = isVisible()
+        val lastSize = size()
+        lastCoordinates = layoutCoordinates
+
+        if (lastVisible != isVisible() || lastSize != size()) {
```
boks1971, line 52: Need some Kotlin education :-) or a new pair of glasses :-)

A couple of lines above these variables (`lastVisible` and `lastSize`) are set to return values from the functions checked against here. Under what conditions will those be not equal?

boks1971 (approved):
lgtm! A couple of small comments/questions.

livekit-android-sdk/src/main/java/io/livekit/android/room/track/video/VideoSinkVisibility.kt
davidliu: the underlying calculations of `isVisible` and `size` depend on the `lastCoordinates`, which gets updated in line 50. They'll be different if the new coordinates are different from the old coordinates.

livekit-android-sdk/src/main/java/io/livekit/android/room/track/Track.kt
davidliu: looks like this one can be cleaned up, good eye.

livekit-android-sdk/src/main/java/io/livekit/android/room/track/video/VideoSinkVisibility.kt
boks1971: Ah got it. Thank you

livekit-android-sdk/src/main/java/io/livekit/android/room/track/video/VideoSinkVisibility.kt
```diff
@@ -0,0 +1,136 @@
...
+class ViewVisibility(private val view: View) : VideoSinkVisibility() {
+
+    private val handler = Handler(Looper.getMainLooper())
+    private val globalLayoutListener = object : ViewTreeObserver.OnGlobalLayoutListener {
+        val lastVisibility = false
+        val lastSize = Track.Dimensions(0, 0)
+
+        override fun onGlobalLayout() {
+            handler.removeCallbacksAndMessages(null)
+            handler.postDelayed({
```
davidzhao, line 75: will this be debouncing the calls? or would it simply queue them up? we'd want a debounced behavior

also, when elements become visible, we should notify immediately to minimize delay of having no content showing. the delay should only be applied to when things are moving to invisible.

davidzhao (commented):
great work!

PR #25, opened 2021-11-24
Compose support

davidliu:


livekit-android-sdk/src/main/java/io/livekit/android/room/Room.kt
```diff
@@ -58,13 +60,24 @@ constructor(
     @Deprecated("Use events instead.")
     var listener: RoomListener? = null
 
-    var sid: Sid? = null
-        private set
-    var name: String? = null
+    /**
+     * Changes can be observed by using [io.livekit.android.util.flow]
+     */
+    var sid: Sid? by flowDelegate(null)
```
davidliu, line 66: These flowDelegates are the guts of the compose support, allowing them to be accessed normally like variables, but also observed using the `flow` function to get notified of changes (and can be used in compose for state purposes).

sample-app-compose/src/main/java/io/livekit/android/composesample/ParticipantItem.kt
```diff
@@ -1,80 +1,116 @@
...
-        return participant
-            .videoTracks.values
-            .firstOrNull()?.track as? RemoteVideoTrack
-    }
 
-    fun setupVideoIfNeeded(videoTrack: RemoteVideoTrack, view: TextureViewRenderer) {
-        if (videoBound) {
-            return
-        }
+    val identity by participant::identity.flow.collectAsState()
```
davidliu, line 34: example of using the flows in compose

davidzhao: this looks neat! how does it know when to refresh?

davidzhao (approved):
looks good!

hiroshihorie:
same with @boks1971 I'm not fluent in Kotlin but will take a look 😂

sample-app-compose/src/main/java/io/livekit/android/composesample/ParticipantItem.kt
davidliu: So kotlin flows are basically an observable stream, and in this case it emits whenever the value is updated. That gets turned into a state variable with the `collectAsState()` method here.

On the compose side, there's just a huge amount of behind the scenes compiler side magic that goes on, but basically whenever a state variable is changed, it triggers a UI update.

davidliu:
> lgtm @davidliu . I really hope someday I will learn Kotlin/Swift etc. to give more useful review feedback :-) Currently, I am just looking for something that kinda looks incorrect :-)

Haha, always good to have more eyes. Feel free to ask any questions on the kotlin stuff, especially since this one's a little crazy on the language specific stuff.

PR #28, opened 2021-12-06
separate room creation from connect

davidliu:


livekit-android-sdk/src/main/java/io/livekit/android/ConnectOptions.kt
```diff
@@ -30,11 +17,6 @@ data class ConnectOptions(
      * capture and publish video track on connect, defaults to false
      */
     val video: Boolean = false,
```
davidliu, line 19: Separated out the defaults and autoManageVideo options from the ConnectOptions into a separate RoomOptions class. These options can be set manually already on the Room object, but seems good to keep them around for exposure purposes. The other ones are related to the actual connection part, so left them in ConnectOptions.

The only ones that could go either way are the audio/video options, since you could have these as a default on the Room object itself, but left them here so it's explicit on connect that it'll create these tracks.

hiroshihorie:
✅ Looks good !, so let's go with separating `ConnectOptions` and `RoomOptions`.

hiroshihorie (approved):
✅ Looks good !, so let's go with separating ConnectOptions and RoomOptions.

livekit-android-sdk/src/main/java/io/livekit/android/ConnectOptions.kt
davidzhao: what about for `autoSubscribe` ? I guess it could go either way.. but I tend to think that's more of a Room option. 

while the docs state "upon connect", I think it's more of a Room level policy. i.e. do I want to subscribe to all tracks in the room, including any new tracks published after joining

davidzhao (commented):
looks good! a friendly reminder to update README as well :)

livekit-android-sdk/src/main/java/io/livekit/android/ConnectOptions.kt
davidliu: Mmm, I agree it *sounds* like a Room option, even though it's only used on connect. However, I think due to the way it behaves currently, it should stay a transient connect only option (or at least unmodifiable), since the client doesn't have a way to modify autoSubscribe behavior after connect.

davidzhao: that makes sense. agreed it shouldn't be changeable. it's ok to be a connect option given the design

PR #29, opened 2021-12-09
handle StreamStateUpdate

davidliu:


livekit-android-sdk/src/main/java/io/livekit/android/room/track/Track.kt
```diff
@@ -12,17 +14,24 @@ open class Track(
...
     val events = eventBus.readOnly()
-
     var name = name
         internal set
+
     var kind = kind
         internal set
     var sid: String? = null
         internal set
+    var streamState: StreamState by flowDelegate(StreamState.ACTIVE) { newValue, oldValue ->
```
boks1971, line 24: @davidliu I am not sure if the client side has an issue with this, i. e. server sends an `ACTIVE` state when it starts streaming a track (in addition to whenever server side pauses and resumes a track). Wondering if this should default to `PAUSED` and fire an event on first `ACTIVE` also

livekit-android-sdk/src/test/java/io/livekit/android/room/SignalClientTest.kt
```diff
@@ -229,6 +229,18 @@ class SignalClientTest {
             build()
         }
 
+        val STREAM_STATE_UPDATE = with(LivekitRtc.SignalResponse.newBuilder()) {
+            streamStateUpdate = with(LivekitRtc.StreamStateUpdate.newBuilder()) {
+                addStreamStates(with(LivekitRtc.StreamStateInfo.newBuilder()) {
+                    participantSid = TestData.REMOTE_PARTICIPANT.sid
+                    trackSid = TestData.REMOTE_AUDIO_TRACK.sid
```
boks1971, line 236: This is test. So, not a big deal. But, just wanted to mention `StreamStateUpdate` is only for video tracks. Audio tracks are not paused/resumed. They are always forwarded

boks1971 (approved):

livekit-android-sdk/src/test/java/io/livekit/android/room/SignalClientTest.kt
davidliu: Good to know!

livekit-android-sdk/src/main/java/io/livekit/android/room/track/Track.kt
boks1971: Yes. It should be. I do not know the details of different client implementations of WebRTC. So, there may be vagaries. But, the mechanism is based on SDP. When an SDP is received from the remote side, clients will call `SetRemoteDescription` on the peer connection. So, whatever tracks are in the SDP, they are available. Pause/Resume on those tracks can happen independently.

By `subscribe`, if you meant exchange with our server, that signaling is actually non-overlapping. The pause can happen only after the client has subscribed. Basically, the full flow is that client subscribes to a track (could be an auto-subscribe too) and there is the SDP part mentioned above exchanged between server and client and tracks are set up. At that point, the server starts forwarding the track. The pipe (i. e. peer connection) used to forward the tracks could get congested and that is when the server needs to pause/resume tracks to avoid congestion. So, pause/resume comes after subscription. If the client unsubscribes, that track is removed from the server side component which is trying to do pause/resume. It will be added back upon re-subscription. So, from client/server signalling protocol `subscribe` sense, there is no overlap of when `subscribe` happens and when `pause/resume` happens. Having said that, it might be possible that there is a `StreamStateUpdate` on a non-existent track (from client side perspective) because of the distributed nature and also implementations. For example, server side might have decided to pause a track and sent down an update. While that message is in transit, client might have unsubscribed from that track and may have removed its track from certain internal structures. When the `StreamStateUpdate` message comes, the corresponding track may not be there. Clients need to be ready for this crossing-on-the-wire behaviour.

davidzhao (commented):
lg!

PR #33, opened 2021-12-17
Fix Subscriber primary data channel

davidliu:


livekit-android-sdk/src/main/java/io/livekit/android/room/RTCEngine.kt
```diff
@@ -179,20 +179,17 @@ internal constructor(
 
         if (joinResponse.subscriberPrimary) {
             // in subscriber primary mode, server side opens sub data channels.
-            publisherObserver.dataChannelListener = onDataChannel@{ dataChannel: DataChannel? ->
-                if (dataChannel == null) {
-                    return@onDataChannel
-                }
+            subscriberObserver.dataChannelListener = onDataChannel@{ dataChannel: DataChannel ->
```
davidliu, line 182: Primary cause of the bug here, was looking for the datachannels on the wrong thing.

boks1971: Are we not using the publisher data channel at all? If the client wants to send data over data channel, is it using the subscriber peer connection data channel to do send that data?

davidliu: The publisher data channel is used for sending data packets, and the subscriber channel is for receiving them.

boks1971 (approved):
lgtm!

PR #35, opened 2021-12-31
handle subscribedqualityupdate

davidliu:


hiroshihorie (approved):
✅

boks1971 (approved):
Looks good @davidliu . Just one request. Server version 0.15.1 was not doing this properly. As that version of code could send this message, we want to ignore this message if it is coming from server version 0.15.1. @davidzhao added a bypass in JS client. Not sure if that same thing can be implemented in mobile clients. If not, we should figure out some way to handle it.

initpwn (approved):

initpwn:
Merge this PR ASAP.

PR #37, opened 2022-01-06
Permissions API

davidliu:
Android seems to have some crashes after repeated enabling/disabling of permissions, but that might be a sample app issue rather than SDK issue. I'll look into that separately. JS seems to handle it fine.

### DIFF
--- a/livekit-android-sdk/src/main/java/io/livekit/android/room/Room.kt
+++ b/livekit-android-sdk/src/main/java/io/livekit/android/room/Room.kt
@@ -226,6 +226,14 @@ constructor(
                             it.streamState
                         )
                     )
+                    is ParticipantEvent.TrackSubscriptionPermissionChanged -> eventBus.postEvent(
+                        RoomEvent.TrackSubscriptionPermissionChanged(
+                            this@Room,
+                            it.participant,
+                            it.trackPublication,
+                            it.subscriptionAllowed
+                        )
+                    )
                 }
             }
         }
@@ -485,21 +493,7 @@ constructor(
 
     override fun onSubscriptionPermissionUpdate(subscriptionPermissionUpdate: LivekitRtc.SubscriptionPermissionUpdate) {
         val participant = getParticipant(subscriptionPermissionUpdate.participantSid) as? RemoteParticipant ?: return
-        val pub = participant.tracks[subscriptionPermissionUpdate.trackSid] as? RemoteTrackPublication ?: return
-
-        if (pub.subscriptionAllowed != subscriptionPermissionUpdate.allowed) {
-            pub.subscriptionAllowed = subscriptionPermissionUpdate.allowed
-
-            // Unsubscribe if become disallowed.
-            if (!pub.subscriptionAllowed && pub.subscribed) {
-                pub.setSubscribed(false)
-            }
-
-            eventBus.postEvent(
-                RoomEvent.TrackSubscriptionPermissionChanged(this, participant, pub, pub.subscriptionAllowed),
-                coroutineScope
-            )
-        }
+        participant.onSubscriptionPermissionUpdate(subscriptionPermissionUpdate)
     }
 
     /**
--- a/livekit-android-sdk/src/main/java/io/livekit/android/room/Room.kt
+++ b/livekit-android-sdk/src/main/java/io/livekit/android/room/Room.kt
@@ -483,14 +483,22 @@ constructor(
         localParticipant.handleSubscribedQualityUpdate(subscribedQualityUpdate)
     }
 
-    override fun onSubscriptionPermissionUpdate(update: LivekitRtc.SubscriptionPermissionUpdate) {
-        val participant = getParticipant(update.participantSid) ?: return
-        val track = participant.tracks[update.trackSid] as? RemoteTrackPublication ?: return
-        track.subscriptionAllowed = update.allowed
+    override fun onSubscriptionPermissionUpdate(subscriptionPermissionUpdate: LivekitRtc.SubscriptionPermissionUpdate) {
+        val participant = getParticipant(subscriptionPermissionUpdate.participantSid) as? RemoteParticipant ?: return
+        val pub = participant.tracks[subscriptionPermissionUpdate.trackSid] as? RemoteTrackPublication ?: return
 
-        // Unsubscribe if become disallowed.
-        if(!track.subscriptionAllowed && track.subscribed) {
-            track.setSubscribed(false)
+        if (pub.subscriptionAllowed != subscriptionPermissionUpdate.allowed) {
+            pub.subscriptionAllowed = subscriptionPermissionUpdate.allowed
+
+            // Unsubscribe if become disallowed.
+            if (!pub.subscriptionAllowed && pub.subscribed) {
+                pub.setSubscribed(false)
+            }
+
+            eventBus.postEvent(
+                RoomEvent.TrackSubscriptionPermissionChanged(this, participant, pub, pub.subscriptionAllowed),
+                coroutineScope
+            )
         }
     }
 
--- a/livekit-android-sdk/src/main/java/io/livekit/android/room/Room.kt
+++ b/livekit-android-sdk/src/main/java/io/livekit/android/room/Room.kt
@@ -483,6 +483,17 @@ constructor(
         localParticipant.handleSubscribedQualityUpdate(subscribedQualityUpdate)
     }
 
+    override fun onSubscriptionPermissionUpdate(update: LivekitRtc.SubscriptionPermissionUpdate) {
+        val participant = getParticipant(update.participantSid) ?: return
+        val track = participant.tracks[update.trackSid] as? RemoteTrackPublication ?: return
+        track.allowed = update.allowed
+
+        // Unsubscribe if become disallowed.
+        if(!track.allowed && track.subscribed) {
+            track.setSubscribed(false)
+        }
+    }
+
     /**
      * @suppress
      */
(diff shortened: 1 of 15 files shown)


livekit-android-sdk/src/main/java/io/livekit/android/room/Room.kt
```diff
@@ -483,6 +483,17 @@ constructor(
         localParticipant.handleSubscribedQualityUpdate(subscribedQualityUpdate)
     }
 
+    override fun onSubscriptionPermissionUpdate(update: LivekitRtc.SubscriptionPermissionUpdate) {
```
boks1971, line 486: Do we want to emit any application facing events when this happens? As a subscriber, you are not able to hear/see a track sounds like a bug. But the reason is because of the source of those tracks not giving you permissions. So, it feels like emitting an event and let the apps notify in whatever way is appropriate to the end user would be helpful.

davidliu:
One question I had was whether tracks should be dropped from the publication when disallowed/unsubscribed? Noticed that we seem not to be doing this at the moment for android/JS.

livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt
```diff
@@ -419,6 +419,30 @@ internal constructor(
...
     }
 
+    /**
+     * Control who can subscribe to LocalParticipant's published tracks.
+     *
+     * By default, all participants can subscribe. This allows fine-grained control over
+     * who is able to subscribe at a participant and track level.
+     *
+     * Note: if access is given at a track-level (i.e. [allParticipantsAllowed] and
+     * [ParticipantTrackPermission.allTracksAllowed] is false), any newer published tracks
```
boks1971, line 429: :nit: `are false`

boks1971: maybe even better to add the word `both` after `i.e.`, something like `i.e. both [allParticipantsAllowed]...` to make it very explicit.

livekit-android-sdk/src/main/java/io/livekit/android/room/track/RemoteTrackPublication.kt
```diff
@@ -88,9 +92,16 @@ class RemoteTrackPublication(
         }
 
     /**
-     * subscribe or unsubscribe from this track
+     * Subscribe or unsubscribe from this track
+     *
+     * If [subscriptionAllowed] is false, subscription will fail.
      */
     fun setSubscribed(subscribed: Boolean) {
+        if (subscribed && !subscriptionAllowed) {
```
boks1971, line 100: Is this preventing sending an `UpdateSubscriptions` message? Sorry, I maybe mistaking code here. But, sending `UpdateSubscriptions` should not be prevented. Server will send the `SubscriptionPermissionUpdate` when there are no permissions. Note that permissions can be revoked when a track is actively subscribed. So, the server could send that update mid stream too. So, clients should always send subscription message and listen for the update from the server.

I think this code is not preventing sending that message, but just wanted to elaborate.

boks1971 (approved):

boks1971:
> One question I had was whether tracks should be dropped from the publication when disallowed/unsubscribed? Noticed that we seem not to be doing this at the moment for android/JS. Looks like we properly get the tracks again through WebRTC when the permissions are regranted.

Yes @davidliu . As permissions can be granted/revoked at any time by the publishing end, server keeps track of state and reconciles the subscriptions with changing permissions and pauses/resumes tracks as necessary. The `SubscriptionPermissionUpdate` message is the mechanism through which server will signal permission changes and whether the associated track will be pausing/resuming. The tracks are not removed from publication.

livekit-android-sdk/src/main/java/io/livekit/android/room/Room.kt
davidliu: Oh yeah, we'll want those, good catch.

davidliu:
> > One question I had was whether tracks should be dropped from the publication when disallowed/unsubscribed? Noticed that we seem not to be doing this at the moment for android/JS. Looks like we properly get the tracks again through WebRTC when the permissions are regranted.
> 
> Yes @davidliu . As permissions can be granted/revoked at any time by the publishing end, server keeps track of state and reconciles the subscriptions with changing permissions and pauses/resumes tracks as necessary. The `SubscriptionPermissionUpdate` message is the mechanism through which server will signal permission changes and whether the associated track will be pausing/resuming. The tracks are not removed from publication.

Sorry, to be clear, is that a yes, we should drop/remove the tracks from the client, or yes, we should keep them around?

livekit-android-sdk/src/main/java/io/livekit/android/room/track/RemoteTrackPublication.kt
boks1971: Got it. Maybe, the log message can be something like `already subscribed, but currently does not have permissions`?

livekit-android-sdk/src/main/java/io/livekit/android/room/track/RemoteTrackPublication.kt
```diff
@@ -63,12 +64,15 @@ class RemoteTrackPublication(
...
+    var subscriptionAllowed: Boolean = true
+        internal set
+
     val isAutoManaged: Boolean
         get() = (track as? RemoteVideoTrack)?.autoManageVideo ?: false
 
     override val subscribed: Boolean
         get() {
-            if (unsubscribed) {
+            if (unsubscribed || !subscriptionAllowed) {
```
boks1971, line 75: Wondering if this should return `false`. The track is subscribed to, but just that it does not have permissions. Feels like this needs to return an enum which gives a better picture of subscription state.

boks1971:
> > > One question I had was whether tracks should be dropped from the publication when disallowed/unsubscribed? Noticed that we seem not to be doing this at the moment for android/JS. Looks like we properly get the tracks again through WebRTC when the permissions are regranted.
> > 
> > 
> > Yes @davidliu . As permissions can be granted/revoked at any time by the publishing end, server keeps track of state and reconciles the subscriptions with changing permissions and pauses/resumes tracks as necessary. The `SubscriptionPermissionUpdate` message is the mechanism through which server will signal permission changes and whether the associated track will be pausing/resuming. The tracks are not removed from publication.
> 
> Sorry, to be clear, is that a yes, we should drop/remove the tracks from the client, or yes, we should keep them around?

Sorry David. I meant, yes - keep the publications. But, I am not saying a definitive yes without knowing all the client intricacies. Basically, the idea is that certain tracks exist in the system and a subscriber has subscribed to it. But those tracks could be paused/resumed based on publisher side permissions. So, if the subscriber checks what tracks are available in the session, the state will say tracks are available, but they may not be pulled down due to permissions. But, I think this is something to discuss with @davidzhao and @hiroshihorie also to find the right approach.

livekit-android-sdk/src/main/java/io/livekit/android/room/track/RemoteTrackPublication.kt
davidzhao: hmm good question, I think technically it's desired to be subscribed, but tricky because clients use `subscribed` to indicate if subscription is successful and ready to be streamed. 

returning false probably makes more sense than true in this case.

livekit-android-sdk/src/main/java/io/livekit/android/room/track/RemoteTrackPublication.kt
davidzhao: I think we should still let the client update the server, because it's indicating a *desired* state.

It does make the API a bit inconsistent, which I don't love:

* .subscribed indicates if it's desired to be subscribed and successfully subscribed
* .setSubscribed updates the desired state

Any alternatives?

livekit-android-sdk/src/main/java/io/livekit/android/room/Room.kt
```diff
@@ -483,6 +483,25 @@ constructor(
...
+    override fun onSubscriptionPermissionUpdate(subscriptionPermissionUpdate: LivekitRtc.SubscriptionPermissionUpdate) {
+        val participant = getParticipant(subscriptionPermissionUpdate.participantSid) as? RemoteParticipant ?: return
+        val pub = participant.tracks[subscriptionPermissionUpdate.trackSid] as? RemoteTrackPublication ?: return
+
+        if (pub.subscriptionAllowed != subscriptionPermissionUpdate.allowed) {
+            pub.subscriptionAllowed = subscriptionPermissionUpdate.allowed
+
+            // Unsubscribe if become disallowed.
+            if (!pub.subscriptionAllowed && pub.subscribed) {
+                pub.setSubscribed(false)
```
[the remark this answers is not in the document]
davidliu: Nope, this was a misunderstanding on my part and has been removed.

livekit-android-sdk/src/main/java/io/livekit/android/room/track/RemoteTrackPublication.kt
davidliu: Discussed with @boks1971, the proposal is to turn `.subscribed` into an enum describing the current subscription status, with "subscribed, not allowed" and "subscribed" as two distinct states.

Thinking about the `setSubscribed` part, I think it might be more descriptive of user intent to have separate `subscribe()` and `unsubscribe()` functions, which would help clear up the inconsistency you note.

livekit-android-sdk/src/main/java/io/livekit/android/room/track/TrackPublication.kt
```diff
@@ -22,10 +22,6 @@ open class TrackPublication(
         private set
     open var muted: Boolean = false
         internal set
-    open val subscribed: Boolean
```
davidzhao, line 25: this is kind of useful when you have a TrackPublication, and don't care if it's remote or local, but just need to quickly check isSubscribed then decide to render.

livekit-android-sdk/src/main/java/io/livekit/android/room/track/RemoteTrackPublication.kt
davidzhao: that sounds good. though I'm wary about breaking current users. Especially on JS. we could perhaps deprecate .subscribed, but still provide it as a way of returning `subscribeState == Subscribed` ?

PR #51, opened 2022-01-31
Send UpdateTrackSettings after resume

davidliu:


livekit-android-sdk/src/main/java/io/livekit/android/room/Room.kt
```diff
@@ -635,8 +636,18 @@ constructor(
...
      */
-    override suspend fun onFullReconnect() {
-        localParticipant.republishTracks()
+    override suspend fun onPostReconnect(isFullReconnect: Boolean) {
+        if (isFullReconnect) {
+            localParticipant.republishTracks()
+        } else {
+            val remoteParticipants = remoteParticipants.values.toList()
+            for (participant in remoteParticipants) {
+                for (pub in participant.videoTracks.values) {
```
boks1971, line 645: `UpdateTrackSettings` has a `disabled` field. I am guessing that is set for audio tracks when subscriber mutes the remote audio track. So, I think this should include update on both audio & video tracks?

boks1971 (approved):

livekit-android-sdk/src/main/java/io/livekit/android/room/Room.kt
davidliu: Hmm, at the moment the server doesn't seem to be using the disabled field for audio tracks. Not sure if we're meaning to use UpdateTrackSettings for audio tracks (though it does make sense for it to.)

boks1971: It should be @davidliu . If not, it is a bug.

It applies to both audio and video (the naming could be confusing) - https://github.com/livekit/livekit-server/blob/745939ec24c32562c7c40dfcbbb9fb53625edab8/pkg/rtc/subscribedtrack.go#L90. It calls `updateDownTrackMute` here (https://github.com/livekit/livekit-server/blob/745939ec24c32562c7c40dfcbbb9fb53625edab8/pkg/rtc/subscribedtrack.go#L101) which will stop the forwarder for that track in the SFU.

livekit-android-sdk/src/main/java/io/livekit/android/room/Room.kt
```diff
@@ -616,8 +617,8 @@ constructor(
     /**
      * @suppress
      */
-    override fun onSignalConnected(isReconnect: Boolean) {
-        if (state == State.RECONNECTING && isReconnect) {
+    override fun onSignalConnected(isFullReconnect: Boolean) {
+        if (state == State.RECONNECTING && isFullReconnect) {
```
davidzhao, line 621: I believe sync state is only needed for migration, not full reconnect.

livekit-android-sdk/src/main/java/io/livekit/android/room/Room.kt
```diff
@@ -635,8 +636,18 @@ constructor(
...
-        localParticipant.republishTracks()
+    override suspend fun onPostReconnect(isFullReconnect: Boolean) {
+        if (isFullReconnect) {
+            localParticipant.republishTracks()
+        } else {
+            val remoteParticipants = remoteParticipants.values.toList()
+            for (participant in remoteParticipants) {
+                for (pub in participant.videoTracks.values) {
+                    val remotePub = pub as? RemoteTrackPublication ?: continue
+                    remotePub.sendUpdateTrackSettings.invoke()
```
davidzhao, line 647: we don't need to send unless it's subscribed.

PR #59, opened 2022-02-21
Queue requests when reconnecting

davidliu:


livekit-android-sdk/src/main/java/io/livekit/android/room/SignalClient.kt
```diff
@@ -564,6 +601,14 @@ constructor(
         const val PROTOCOL_VERSION = 6
         const val SDK_TYPE = "android"
 
+        private val skipQueueTypes = listOf(
+            LivekitRtc.SignalRequest.MessageCase.SYNC_STATE,
+            LivekitRtc.SignalRequest.MessageCase.TRICKLE,
+            LivekitRtc.SignalRequest.MessageCase.OFFER,
+            LivekitRtc.SignalRequest.MessageCase.ANSWER,
+            LivekitRtc.SignalRequest.MessageCase.SIMULATE
+        )
```
hiroshihorie, line 610: Nice! so these are the types we should skip ?

hiroshihorie (approved):

PR #67, opened 2022-03-23
Fix simulcast factory not sending back EncoderInfo

davidliu:


livekit-android-sdk/src/main/java/io/livekit/android/webrtc/SimulcastVideoEncoderFactoryWrapper.kt
```diff
@@ -165,6 +168,31 @@ internal class SimulcastVideoEncoderFactoryWrapper(
...
+            return future.get()
+        }
+
+        override fun getResolutionBitrateLimits(): Array<VideoEncoder.ResolutionBitrateLimits> {
+            val future = executor.submit(Callable { return@Callable encoder.resolutionBitrateLimits })
+            return future.get()
+        }
+
+        override fun getEncoderInfo(): VideoEncoder.EncoderInfo {
+            val future = executor.submit(Callable { return@Callable encoder.encoderInfo })
```
davidliu, line 193: This is the main thing that was breaking the hardware encoder from working (and preventing H264 from working). EncoderInfo holds onto information needed for resolution alignment, which is used to adjust the resolution (only works on 16x16 aligned frames).

hiroshihorie (approved):
Nice fix !

davidzhao:
would flutter have the same issue then?

davidliu:
Very likely, I'll put in a PR for Flutter tomorrow.

It looks like this behavior got broken in the M97 upgrade, so the device farm testing will hopefully help catch these breakages down the road.

PR #85, opened 2022-05-14
Compose visibility fixes

davidliu:


livekit-android-sdk/src/main/java/io/livekit/android/room/track/RemoteTrackPublication.kt
```diff
@@ -40,7 +40,7 @@ class RemoteTrackPublication(
                     }
                 }
 
-                if (value is RemoteVideoTrack) {
+                if (value is RemoteVideoTrack && value.autoManageVideo) {
```
davidzhao, line 43: can we change `autoManageVideo` to `adaptiveStream` within internal vars too?

PR #100, opened 2022-06-24
handle audio output changes automatically in sample apps

davidliu:


davidzhao:
this is great! Should we include some level of this capability within the SDK?

davidliu:
For now, thinking not, since the code is taken from the WebRTC demo and might be a little outdated. Twilio actually has a separate library for this that people can use, would need to invest quite a bit of time to get up to par.

PR #104, opened 2022-06-29
Have SDK handle audio routing by default

davidliu:
Use AudioSwitch library for this rather than owning the audio routing code for now.


davidzhao:
nice!

PR #108, opened 2022-07-05
Permanent local participant to fix NPEs when not connected

davidliu:


hiroshihorie (approved):
Cool, so now `localParticipant` is a `LocalParticipant` instead of a `LocalParticipant?` .

davidliu:
> Cool, so now `localParticipant` is a `LocalParticipant` instead of a `LocalParticipant?` .

Yeah.

PR #153, opened 2022-10-21
Add capture observer factory to LocalVideoTrackOptions

davidliu:
Allows for developers to receive the frames that are being recorded by the video capturer.


davidzhao (approved):
this looks great! it would be great to have an example app that saves a user's audio and video tracks locally while publishing. primarily to show users how to interact with the encoding stack.

davidliu:
Was planning a wiki page for this, but yeah, an example app makes a lot of sense.

davidliu:
Realized that this isn't really needed, can actually just attach to VideoTracks as a sink and get frames that way.

Issue #179, opened 2023-02-09
camera failed in some devices


dat0106:
**Describe the bug**
camera in some devices (Redmi note 11pro, Redmi note 11, one plus) does not show in other device,

**Expected behavior**

**Screenshots**
![20230209_163503 (1)](https://user-images.githubusercontent.com/4971539/217777527-cbcf771b-5984-4f3c-8052-d1ee06bf2b87.jpg)

**Device Info:**
 - Device: Redmi note 11pro
 - OS:  Android 12
 - LiveKit SDK version: 1.1.6

Test-isom:
I'am experiencing same issue on redmi note 7 device and changing version of protobuf  to "3.21.12" solves it for me, can you please release a new version by merging this fix if possible @davidliu

Test-isom:
Yeah i was going through his comment and even saw another videos he uploaded on youtube for issue, but for now i added these lines in build.gradle along with the livekit sdk dependency
and since grade would take latest version of dependency if there are conflicting version so it works for me for now on redmi note devices.

//For https://github.com/livekit/client-sdk-android/issues/179
implementation "com.google.protobuf:protobuf-javalite:3.21.12"

But thanks a lot for the reply, will be actively following the repo for a better fix,until than this works for me so i guess no need for a updated sdk.

davidliu:
Protobuf updated to 3.22.0 in #186.

lihongyang166:
@MikuMikuQAQ   Thank you very much, it has been solved！

davidliu:
Hmm, looks like there's some issues here, perhaps with either the simulcast encoding or software encoding. I'll dig into this a little more and see if we can fix it so that the overrides aren't needed.

davidliu:
Looks like this device has issues with H264 simulcast, and then switches to VP8 encoding on the fly. At the moment, the server doesn't handle these switches well, and ends up failing.

wzJun1:
[...]

PR #180, opened 2023-02-09
fix bug :https://github.com/livekit/client-sdk-android/issues/179

dat0106:
upgrade lib protobuf

### DIFF
--- a/deps.gradle
+++ b/deps.gradle
@@ -19,7 +19,7 @@ ext {
             junitJupiter      : "5.5.0",
             coroutines        : "1.6.0",
             lint              : "30.0.1",
-            protobuf          : "3.19.4",
+            protobuf          : "3.21.12",
     ]
     generated = [
             protoSrc: "$projectDir/protocol",


davidliu:
Does this fix the camera issue for you?

davidliu:
Please sign the CLA link: https://cla-assistant.io/livekit/client-sdk-android?pullRequest=180

davidliu:
Do you have any logs which pointed at protobuf being the issue here? Trying to figure out how this fixes the issue here.

dat0106:
I tried more testing, but still getting error, maybe it's not the problem, let me do more research on this
https://youtube.com/shorts/troOMl3YElc

lihongyang166:
There is a situation that Huawei mobile phone and tablet can not push the stream when using the camera. The system uses Hongmeng 3.0 compatible with the Android system. The current situation is that there is only audio but no video. There is no problem when testing Xiaomi mobile phones. How to solve this problem?

PR #199, opened 2023-03-14
Switch to webrtc hosted on maven central

davidliu:


davidzhao (approved):
Nice! Mind updating the readme too? I think we can drop gitpack now right?

PR #270, opened 2023-09-15
Add spotless check to CI builds

davidliu:


davidliu:
Build expected to fail here, there's a couple third party licenses here which won't pass the spotless check, but they shouldn't fail in the future since it's racheted to main, and we don't touch these files often.

PR #276, opened 2023-09-19
Improvements for E2EE.

cloudwebrtc:
* add more `KeyProvider` API
* use `participant.identity` as the `participantId` for FrameCryptor.
* set server injection frame trailer from joinResponse
* remove Framcryptor when unpublished/unsubscribed

### DIFF
--- a/livekit-android-sdk/src/main/java/io/livekit/android/e2ee/E2EEManager.kt
+++ b/livekit-android-sdk/src/main/java/io/livekit/android/e2ee/E2EEManager.kt
@@ -169,7 +169,7 @@ constructor(keyProvider: KeyProvider) {
             keyProvider.rtcKeyProvider,
         )
 
-        frameCryptors[HashMap<String, String>().apply {
+        frameCryptors[trackId to participantId] = frameCryptor
             put(trackId, participantId)
         }] = frameCryptor
         frameCryptor.setEnabled(enabled)
--- a/livekit-android-sdk/src/main/java/io/livekit/android/e2ee/E2EEManager.kt
+++ b/livekit-android-sdk/src/main/java/io/livekit/android/e2ee/E2EEManager.kt
@@ -99,17 +99,13 @@ constructor(keyProvider: KeyProvider) {
     }
 
     public fun removeSubscribedTrack(track: Track, publication: TrackPublication, participant: RemoteParticipant, room: Room) {
-        for (item in frameCryptors.entries) {
-            var key = item.key
-            var trackId = publication.sid
-            var participantId = participant.identity
-            if (key.first == trackId && key.second == participantId) {
-                var frameCryptor = item.value
-                frameCryptor.isEnabled = false
-                frameCryptor.dispose()
-                frameCryptors.remove(item.key)
-                return
-            }
+        var trackId = publication.sid
+        var participantId = participant.identity
+        var frameCryptor = frameCryptors.get(trackId to participantId)
+        if (frameCryptor != null) {
+            frameCryptor.isEnabled = false
+            frameCryptor.dispose()
+            frameCryptors.remove(trackId to participantId)
         }
     }
 
@@ -138,17 +134,13 @@ constructor(keyProvider: KeyProvider) {
     }
 
     public fun removePublishedTrack(track: Track, publication: TrackPublication, participant: LocalParticipant, room: Room) {
-        for (item in frameCryptors.entries) {
-            var key = item.key
-            var trackId = publication.sid
-            var participantId = participant.identity
-            if (key.first == trackId && key.second == participantId) {
-                var frameCryptor = item.value
-                frameCryptor.isEnabled = false
-                frameCryptor.dispose()
-                frameCryptors.remove(item.key)
-                return
-            }
+        var trackId = publication.sid
+        var participantId = participant.identity
+        var frameCryptor = frameCryptors.get(trackId to participantId)
+        if (frameCryptor != null) {
+            frameCryptor.isEnabled = false
+            frameCryptor.dispose()
+            frameCryptors.remove(trackId to participantId)
         }
     }
 
--- a/livekit-android-sdk/src/main/java/io/livekit/android/e2ee/E2EEManager.kt
+++ b/livekit-android-sdk/src/main/java/io/livekit/android/e2ee/E2EEManager.kt
@@ -37,7 +37,7 @@ class E2EEManager
 constructor(keyProvider: KeyProvider) {
     private var room: Room? = null
     private var keyProvider: KeyProvider
-    private var frameCryptors = mutableMapOf<Map<String, String>, FrameCryptor>()
+    private var frameCryptors = mutableMapOf<Pair<String, String>, FrameCryptor>()
     private var algorithm: FrameCryptorAlgorithm = FrameCryptorAlgorithm.AES_GCM
     private lateinit var emitEvent: (roomEvent: RoomEvent) -> Unit?
     var enabled: Boolean = false
--- a/livekit-android-sdk/src/main/java/io/livekit/android/e2ee/E2EEManager.kt
+++ b/livekit-android-sdk/src/main/java/io/livekit/android/e2ee/E2EEManager.kt
@@ -37,12 +37,16 @@ class E2EEManager
 constructor(keyProvider: KeyProvider) {
     private var room: Room? = null
     private var keyProvider: KeyProvider
-    private var frameCryptors = mutableMapOf<String, FrameCryptor>()
+    private var frameCryptors = mutableMapOf<Map<String, String>, FrameCryptor>()
     private var algorithm: FrameCryptorAlgorithm = FrameCryptorAlgorithm.AES_GCM
     private lateinit var emitEvent: (roomEvent: RoomEvent) -> Unit?
     var enabled: Boolean = false
     init {
         this.keyProvider = keyProvider
+    }
+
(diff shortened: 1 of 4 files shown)


livekit-android-sdk/src/main/java/io/livekit/android/e2ee/E2EEManager.kt
```diff
@@ -136,41 +162,32 @@ constructor(keyProvider: KeyProvider) {
...
         var frameCryptor = FrameCryptorFactory.createFrameCryptorForRtpReceiver(
             receiver,
-            pid,
+            participantId,
             algorithm,
             keyProvider.rtcKeyProvider,
         )
 
-        frameCryptors[trackId] = frameCryptor
+        frameCryptors[HashMap<String, String>().apply {
```
davidliu, line 187:
```suggestion
        frameCryptors[mapOf(trackId to participantId)]
```

livekit-android-sdk/src/main/java/io/livekit/android/e2ee/E2EEManager.kt
```diff
@@ -37,14 +37,18 @@ class E2EEManager
 constructor(keyProvider: KeyProvider) {
     private var room: Room? = null
     private var keyProvider: KeyProvider
-    private var frameCryptors = mutableMapOf<String, FrameCryptor>()
+    private var frameCryptors = mutableMapOf<Map<String, String>, FrameCryptor>()
```
davidliu, line 40:
```suggestion
    private var frameCryptors = mutableMapOf<Pair<String, String>, FrameCryptor>()
```

Looks like the key of frameCryptors should just be an immutable pair rather than a full map?

livekit-android-sdk/src/main/java/io/livekit/android/e2ee/E2EEManager.kt
```diff
@@ -136,41 +162,32 @@ constructor(keyProvider: KeyProvider) {
...
         var frameCryptor = FrameCryptorFactory.createFrameCryptorForRtpSender(
             sender,
-            pid,
+            participantId,
             algorithm,
             keyProvider.rtcKeyProvider,
         )
 
-        frameCryptors[trackId] = frameCryptor
+        frameCryptors[HashMap<String, String>().apply {
```
davidliu, line 172:
```suggestion
        frameCryptors[trackId to participantId] = frameCryptor
```

livekit-android-sdk/src/main/java/io/livekit/android/e2ee/E2EEManager.kt
davidliu, line 187:
```suggestion
        frameCryptors[trackId to participantId] = frameCryptor
```

livekit-android-sdk/src/main/java/io/livekit/android/e2ee/E2EEManager.kt
```diff
@@ -96,9 +98,20 @@ constructor(keyProvider: KeyProvider) {
         }
     }
 
+    public fun removeSubscribedTrack(track: Track, publication: TrackPublication, participant: RemoteParticipant, room: Room) {
+        for (item in frameCryptors.entries) {
+            var key = item.key
+            var frameCryptor = item.value
+            if (key[publication.sid] == participant.identity) {
```
davidliu, line 105:
```suggestion
            if (key.second == participant.identity) {
```

livekit-android-sdk/src/main/java/io/livekit/android/e2ee/E2EEManager.kt
```diff
@@ -122,6 +135,19 @@ constructor(keyProvider: KeyProvider) {
         }
     }
 
+    public fun removePublishedTrack(track: Track, publication: TrackPublication, participant: LocalParticipant, room: Room) {
+        for (item in frameCryptors.entries) {
+            var key = item.key
+            var frameCryptor = item.value
+            if (key[publication.sid] == participant.identity) {
```
davidliu, line 142:
```suggestion
            if (key.second == participant.identity) {
```

livekit-android-sdk/src/main/java/io/livekit/android/room/Room.kt
```diff
@@ -973,6 +980,9 @@ constructor(
         participant: RemoteParticipant,
     ) {
         listener?.onTrackUnsubscribed(track, publication, participant, this)
+        if (e2eeManager != null) {
```
davidliu, line 983:
```suggestion
        e2eeManager?.let { e2eeManager ->
```

Removes the need for the `!!` in the following line.

livekit-android-sdk/src/main/java/io/livekit/android/room/Room.kt
```diff
@@ -938,6 +942,9 @@ constructor(
      */
     override fun onTrackUnpublished(publication: LocalTrackPublication, participant: LocalParticipant) {
         listener?.onTrackUnpublished(publication, participant, this)
+        if (e2eeManager != null) {
```
davidliu, line 945:
```suggestion
        e2eeManager?.let { e2eeManager ->
```

Removes the need for the `!!` in the following line.

livekit-android-sdk/src/main/java/io/livekit/android/e2ee/E2EEManager.kt
```diff
@@ -96,9 +98,20 @@ constructor(keyProvider: KeyProvider) {
         }
     }
 
+    public fun removeSubscribedTrack(track: Track, publication: TrackPublication, participant: RemoteParticipant, room: Room) {
+        for (item in frameCryptors.entries) {
+            var key = item.key
+            var frameCryptor = item.value
+            if (key.first == participant.sid) {
```
davidliu, line 105: I think this should be `key.second`? trackId is in the first position and participantId in the second position.

livekit-android-sdk/src/main/java/io/livekit/android/e2ee/E2EEManager.kt
```diff
@@ -122,6 +135,19 @@ constructor(keyProvider: KeyProvider) {
         }
     }
 
+    public fun removePublishedTrack(track: Track, publication: TrackPublication, participant: LocalParticipant, room: Room) {
+        for (item in frameCryptors.entries) {
+            var key = item.key
+            var frameCryptor = item.value
+            if (key.first == participant.sid) {
```
davidliu, line 142: I think this should be `key.second`? trackId is in the first position and participantId in the second position.

livekit-android-sdk/src/main/java/io/livekit/android/e2ee/E2EEManager.kt
cloudwebrtc: I think sid should be used here because it is unique and corresponds to the actual audio or video track. If the participant ID is used to remove the framecryptor, the video framecryptor may also be removed when the other party just unpublish the audio track, causing the video decoding to fail.

davidliu: I mean that you put `trackId to participantId` here:

https://github.com/livekit/client-sdk-android/pull/276/files#diff-6dbe0ea2e715879edad961dea930389cbd5bd274e22b5c7991d7649d1f6f88dfR172

This means `key.first` equals trackId, and `key.second` equals participantId. The if statement here would never pass since you're checking a track id against a participant id.

cloudwebrtc: maybe we just need trackId/sid as the frameCryptors Map key, or change to 
`if (key.first == participant.sid && key.second == publication.sid) {`

livekit-android-sdk/src/main/java/io/livekit/android/e2ee/E2EEManager.kt
cloudwebrtc: okay, I found the error, this should be `publication.sid` not `participant.sid`, which is being verified.

livekit-android-sdk/src/main/java/io/livekit/android/e2ee/E2EEManager.kt
cloudwebrtc: I've fixed it and testing works fine

livekit-android-sdk/src/main/java/io/livekit/android/e2ee/E2EEManager.kt
```diff
@@ -122,6 +137,21 @@ constructor(keyProvider: KeyProvider) {
         }
     }
 
+    public fun removePublishedTrack(track: Track, publication: TrackPublication, participant: LocalParticipant, room: Room) {
+        for (item in frameCryptors.entries) {
+            var key = item.key
+            var trackId = publication.sid
+            var participantId = participant.identity
+            if (key.first == trackId && key.second == participantId) {
```
davidliu, line 145: If we're checking against both parts of the pair, can just do `frameCryptors[trackId to participantId]` to grab the frameCryptor if any.

cloudwebrtc: How does this look like, https://github.com/livekit/client-sdk-android/pull/276/commits/d9fc42e0fc7a6654fe0ddb52f4097f3ae2eb71e4

davidliu (approved):

PR #301, opened 2023-10-26
Fix: Sometimes, the reconnection logic can hang on a flaky internet connection.

dshkil:
The root cause is obvious: `reconnectingLock` was not released when `this.isClosed == true`.
To fix the issue, we can handle the case without using the `reconnectingLock`.


davidliu:
Hmm, I don't want to remove the mutex here (there's going to be more mutex use-cases within the RTCEngine). Can you just add unlock statement in the fail-fast cases?

davidliu (approved):

PR #304, opened 2023-11-01
SVC Codec support

davidliu:


livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt
```diff
@@ -267,9 +272,23 @@ internal constructor(
...
     ) {
+        val isSVC = isSVCCodec(options.videoCodec)
+
+        @Suppress("NAME_SHADOWING") var options = options
+        if (isSVC) {
+            dynacast = true
+
+            // Ensure backup codec and scalability for svc codecs.
+            if (options.backupCodec == null) {
+                options = options.copy(backupCodec = BackupVideoCodec())
```
davidzhao, line 283: I guess this would enable multicodec automatically if null?

davidzhao: I do like that default, but feels like it would be a bit inconsistent if there isn't a way to turn it off.

JS currently allows users to set it to `true` to enable with defaults: https://github.com/livekit/client-sdk-js/pull/923

livekit-android-sdk/src/main/java/io/livekit/android/room/participant/PublishUtils.kt
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.livekit.android.room.participant
+
+fun String.mimeTypeToVideoCodec(): String {
+    return split("/")[1].lowercase()
```
davidzhao, line 20: need to guard for unexpected response?

davidzhao (commented):
LGTM!

PR #312, opened 2023-11-07
Guard against malformed server response and fix server codec handling

davidliu:


livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt
```diff
@@ -850,6 +850,13 @@ abstract class BaseVideoTrackPublishOptions {
...
     abstract val scalabilityMode: String?
 
+    /**
+     * Multi-codec Simulcast
+     *
+     * Codecs such as VP9 and AV1 are not supported by all clients. When backupCodec is
+     * set, when an incompatible client attempts to subscribe to the track, LiveKit
+     * will automatically publish a secondary track encoded with the backup codec.
+     */
     abstract val backupCodec: BackupVideoCodec?
```
davidzhao, line 860: is this default enabled on Android? or something you'd need to set?

davidliu: Always on for svc codecs atm, will need to address being able to turn it off for svc codecs. Otherwise empty for other codecs.

davidzhao: cool, yeah let's discuss default behavior, ideally it's consistent across SDKs

PR #325, opened 2023-11-20
Fixed memory leak - LeakCanary leak detection io.livekit.android.rend…

### LINKED ISSUE #324: LeakCanary leak detection io.livekit.android.renderer.TextureViewRenderer
### Description
After conducting a video call in our application, we observed memory leaks being logged. The leak seems to be related to `TextureViewRenderer` instances, which are holding references to a destroyed activity. This issue was detected using LeakCanary.

### LeakCanary Log

`
1 APPLICATION LEAKS

References underlined with "~~~" are likely causes.
Learn more at https://squ.re/leaks.

175409 bytes retained by leaking objects
Signature: da39a3ee5e6b4b0d3255bfef95601890afd80709
┬───
│ GC Root: Global variable in native code
│
├─ io.livekit.android.renderer.TextureViewRenderer instance
│ Leaking: YES (View.mContext references a destroyed activity)
│ Retaining 175.4 kB in 3367 objects
│ View not part of a window view hierarchy
│ View.mAttachInfo is null (view detached)
│ View.mWindowAttachCount = 1
│ mContext instance of [REDACTED].activities.CallActivity with mDestroyed = true
│ ↓ View.mContext
╰→ [REDACTED].activities.CallActivity instance
​ Leaking: YES (ObjectWatcher was watching this because [REDACTED].activities.CallActivity received
​ Activity#onDestroy() callback and Activity#mDestroyed is true)
​ Retaining 170.3 kB in 3321 objects
​ key = c5623187-0999-484e-bb56-5dc03dfebe9f
​ watchDurationMillis = 5285
​ retainedDurationMillis = 285
​ mApplication instance of [REDACTED].Application
​ mBase instance of android.app.ContextImpl`

serhiynovos:
I noticed that sometimes when my app leaves a call, LeakCanary shows that memory leak is detected. After these changes I stopped seeing this message

This PR should fix #324 issue

### DIFF
--- a/livekit-android-sdk/src/main/java/io/livekit/android/compose/VideoRenderer.kt
+++ b/livekit-android-sdk/src/main/java/io/livekit/android/compose/VideoRenderer.kt
@@ -83,7 +83,12 @@ fun VideoRenderer(
 
     DisposableEffect(currentCompositeKeyHash.toString()) {
         onDispose {
-            view?.release()
+            view?.let {
+                it.release()
+                videoTrack?.removeRenderer(it)
+            }
+            view = null
+            boundVideoTrack = null
         }
     }
 


serrg1994:
@davidliu I'm thinking if it could not be issue of random crashes on peerConnection dispose as looks like it had references to tracks and some native memory may leak which could case that something wanted to access peerconneection which was not in memory anymore

livekit-android-sdk/src/main/java/io/livekit/android/compose/VideoRenderer.kt
```diff
@@ -83,7 +83,12 @@ fun VideoRenderer(
 
     DisposableEffect(currentCompositeKeyHash.toString()) {
         onDispose {
-            view?.release()
+            view?.let {
+                it.release()
+                videoTrack?.removeRenderer(it)
```
davidliu, line 88: Hmm, is the `cleanupVideoTrack()` not getting called for you? That's essentially what this is doing, so it seems redundant.

serhiynovos: I'll test more. But only after these changes I stopped getting leak notification

PR #345, opened 2023-12-20
Allow registration of external cameras implementations

jfilo-ebay:
Android Camera support still evolves while propagating changes to downstream libraries might be overwhelming we at eBay suggest leaving a door open for consumer's integration of Camera.

This change modifies behaviour in CameraCapturerUtils to allow external definitions of Camera provided by new interface of CameraProvider based on WebRTC interfaces and implementations.


davidliu (approved):
Some small cleanup stuff, but LGTM!

PR #350, opened 2024-01-07
Allow set connect retry to connect function

mohsen2986:
In certain scenarios, users may encounter connectivity issues that prevent them from joining a room. Implementing a simple retry mechanism can effectively resolve these issues.


livekit-android-sdk/src/main/java/io/livekit/android/LiveKit.kt
```diff
@@ -132,7 +135,10 @@ class LiveKit {
             val protocolVersion = maxOf(options.protocolVersion, ProtocolVersion.v7)
             val connectOptions = options.copy(protocolVersion = protocolVersion)
 
-            room.connect(url, token, connectOptions)
```
davidliu, line 135: This should be added to the room.connect function itself rather than this one. `LiveKit.connect` is deprecated and will be removed in the future.

PR #389, opened 2024-03-07
Fix network request leak on pre-8.1 devices

davidliu:


davidliu:
For future reference: https://android.googlesource.com/platform/frameworks/base/+/dafed3d00818713b60ec09cfff482487a3405202

Pre 8.1 didn't have this fix, and so `unregisterNetworkCallback` would only unregister the most recent request and "leak" prior requests. An app can only have up to 100 ongoing requests at a time, otherwise it'll throw an exception. Notably, WebRTC also registers a callback in `NetworkMonitorAutoDetect`, so the exception would crash the app.

PR #392, opened 2024-03-13
add foregroundServiceType for ForegroundService

jossephus:
To run the sample app in Android 14, ForegroundServiceType should be specified for foreground services: [Documentation LINK](https://developer.android.com/about/versions/14/changes/fgs-types-required)


davidliu:
Shoot, forgot about this Android 14 change. Thanks for the PR!

PR #435, opened 2024-06-14
Fix setScreenShareEnabled not starting the screen share track

davidliu:


DaniilKharin32:
Please bring back the function of displaying a custom screencast notification

PR #544, opened 2024-11-19
Fix surface causing null pointer exception on some devices

KasemJaffer:
Steps to reproduce. Turn camera on and off then turn it back on. It crashes with null pointer exception.

### DIFF
--- a/livekit-android-camerax/src/main/java/livekit/org/webrtc/CameraXSession.kt
+++ b/livekit-android-camerax/src/main/java/livekit/org/webrtc/CameraXSession.kt
@@ -147,7 +147,9 @@ internal constructor(
 
                 surface = Surface(surfaceTextureHelper.surfaceTexture)
                 surfaceProvider = SurfaceProvider { request ->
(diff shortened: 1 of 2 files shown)


livekit-android-camerax/src/main/java/livekit/org/webrtc/CameraXSession.kt
```diff
@@ -147,7 +147,7 @@ internal constructor(
 
                 surface = Surface(surfaceTextureHelper.surfaceTexture)
                 surfaceProvider = SurfaceProvider { request ->
-                    request.provideSurface(surface!!, helperExecutor) { }
```
davidliu, line 150: Surface is only nulled out when the camera is stopped, so we'll need to call `request.willNotComplete()` if we're not providing it any more.

https://developer.android.com/reference/androidx/camera/core/Preview.SurfaceProvider#onSurfaceRequested(androidx.camera.core.SurfaceRequest)

davidliu (approved):

PR #546, opened 2024-11-21
Fix NoSuchMethodError when calling CameraXHelper.createCameraProvider without useCases

KasemJaffer:
This fixes an exception that gets thrown in runtime (for some reason) if you try to call `CameraXHelper.createCameraProvider` without the optional `useCases` argument.

`
FATAL EXCEPTION: main Process: io.livekit.android, PID: 30063 java.lang.NoSuchMethodError: No virtual method createCameraProvider(Landroidx/lifecycle/LifecycleOwner;)Lio/livekit/android/room/track/video/CameraCapturerUtils$CameraProvider; in class Llivekit/org/webrtc/CameraXHelper$Companion; or its super classes
`

### DIFF
--- a/livekit-android-camerax/src/main/java/livekit/org/webrtc/CameraXHelper.kt
+++ b/livekit-android-camerax/src/main/java/livekit/org/webrtc/CameraXHelper.kt
@@ -29,19 +29,6 @@ import io.livekit.android.room.track.video.CameraEventsDispatchHandler
 
 class CameraXHelper {
     companion object {
-
-        /**
-         * Creates a CameraProvider that uses CameraX for its sessions.
-         *
-         * For use with [CameraCapturerUtils.registerCameraProvider].
-         * Remember to unregister the provider when outside the lifecycle
-         * of [lifecycleOwner].
-         *
-         * @param lifecycleOwner The lifecycleOwner which controls the lifecycle transitions of the use cases.
-         */
-        @ExperimentalCamera2Interop
-        fun createCameraProvider(lifecycleOwner: LifecycleOwner) = createCameraProvider(lifecycleOwner, emptyArray())
-
         /**
          * Creates a CameraProvider that uses CameraX for its sessions.
          *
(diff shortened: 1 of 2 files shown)


KasemJaffer:
@davidliu I'm not sure if i need to create release changeSet for this. It's a patch for a yet to be released version.

davidliu:
> @davidliu I'm not sure if i need to create release changeSet for this. It's a patch for a yet to be released version.

Nah, won't need a changeset for this.

davidliu:
> This fixes an exception that gets thrown in runtime (for some reason) if you try to call `CameraXHelper.createCameraProvider` without the optional `useCases` argument.
> 
> `FATAL EXCEPTION: main Process: io.livekit.android, PID: 30063 java.lang.NoSuchMethodError: No virtual method createCameraProvider(Landroidx/lifecycle/LifecycleOwner;)Lio/livekit/android/room/track/video/CameraCapturerUtils$CameraProvider; in class Llivekit/org/webrtc/CameraXHelper$Companion; or its super classes`

Hmm, I'm guessing the original code was linked against the old version, and then the dependency upgraded somewhere else in the tree, changing the method signature in the final product.

livekit-android-camerax/src/main/java/livekit/org/webrtc/CameraXHelper.kt
```diff
@@ -30,6 +30,18 @@ import io.livekit.android.room.track.video.CameraEventsDispatchHandler
...
+        /**
+         * Creates a CameraProvider that uses CameraX for its sessions.
+         *
+         * For use with [CameraCapturerUtils.registerCameraProvider].
+         * Remember to unregister the provider when outside the lifecycle
+         * of [lifecycleOwner].
+         *
+         * @param lifecycleOwner The lifecycleOwner which controls the lifecycle transitions of the use cases.
+         */
+        @ExperimentalCamera2Interop
```
davidliu, line 42: I think adding `@JvmOverloads` to the original method would solve this. Would be a little cleaner that way.

Issue #535, opened 2024-11-08
Rotation of published video is always portrait while in background


bennnjamin:
**Describe the bug**
When testing out the sample apps, you can rotate the device to switch between publishing landscape and portrait video. If you bind the Camera to a service lifecycle so that you can publish video in the call while in the background then upon pressing the home button the video will revert back to portrait.

**To Reproduce**
1. Enable background video publishing using a Foreground Service
2. Publish video to a room
3. Rotate device to landscape to publish landscape video
4. Press home button
5. Video reverts back to portrait

**Expected behavior**
The video rotation should match the device, or the SDK should expose APIs so the developer can manage video rotation.

**Device Info:**
 - Device: Samsung S22
 - OS: Android 14
 - LiveKit SDK version: 2.10.0

**Additional context**
- I've tried forcing the app to only run in landscape but that doesn't fix the issue
- I've tried both with and without CameraX helper
- I have looked for methods in the SDK to control rotation manually and haven't found any

bennnjamin:
I have found that the issue is this method in the WebRTC fork https://github.com/webrtc-sdk/webrtc/blob/543121ba1cd47780e92d48546b880333265b37b5/sdk/android/src/java/org/webrtc/CameraSession.java#L42

I can close and re-open this issue there but it is also possible to workaround the issue in this repo in the `getFrameOrientation` method of `CameraXSession`.

davidliu:
Tried it myself, and it seems to follow the orientation of the device's UI rotation (which may not physically line up with actual rotation). That is, since the home launcher doesn't support landscape (on my device at least), it forces the UI rotation to portrait and the camera follows that. If you go into a separate app that does support landscape, the camera will correctly reorient to landscape as well.

Closing as intended behavior, since users aren't likely to stay in the landscape physical rotation when entering a portrait only app.

PR #552, opened 2024-12-09
Detect screenshare rotation

davidliu:


bennnjamin:
These changes look very similar to what is required to fix this issue #535. Especially since it's possible for the device to be in landscape while screensharing when switching between apps. I would greatly appreciate if you took another look at a robust orientation handling solution for the camera in addition to screensharing.

davidliu:
@bennnjamin As I noted in #535, the camera is already following the rotation of the device's UI, so I'm not sure what's actionable there.

bennnjamin:
@davidliu it doesn't follow the device's UI when switching apps or when there is no foreground activity. It's certainly jarring to a subscriber to see the camera swapping between landscape and portrait orientation during normal operation of the phone. The expectation from a UX perspective is that the video orientation matches the device even when switching apps or multitasking.

The action is to use ```OrientationListener``` instead of ```WindowManager``` for camera video track orientation like in this pull request but even better would be to allow the developer override the frame orientation. Then it's up to them if they would like to override the default behavior with a custom `OrientationListener`.

davidliu:
@bennnjamin that's possible to do with a VideoProcessor, though you shouldn't have to. If you can provide a repro example, I'll gladly recheck it and provide a fix if needed, but it was working fine for me.

bennnjamin:
@davidliu I appreciate your willingness to help with this. Perhaps I was looking in the wrong place to override rotation. Can you link to a method that is overridable? Here is are the methods I found for frame rotation but they are all private.

https://github.com/livekit/client-sdk-android/blob/dcbb36b25a318c53c5d15a27f4e83fd8e73b2da3/livekit-android-camerax/src/main/java/livekit/org/webrtc/CameraXSession.kt#L305

And in WebRTC:

https://github.com/webrtc-sdk/webrtc/blob/b99fd2c270361aea2d458e61ac4a4cd2443bdbf6/sdk/android/src/java/org/webrtc/Camera2Session.java#L413


https://github.com/webrtc-sdk/webrtc/blob/b99fd2c270361aea2d458e61ac4a4cd2443bdbf6/sdk/android/src/java/org/webrtc/CameraSession.java#L42

bennnjamin:
@davidliu kindly checking to see if you are able to provide further guidance on the correct place to override frame rotation for `Camera2` and `Camerax` sessions

PR #578, opened 2025-01-08
Implement RPC

davidliu:


livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt
```diff
@@ -1232,3 +1608,23 @@ internal fun VideoTrackPublishOptions.hasBackupCodec(): Boolean {
...
+
+/**
+ * A handler that processes an RPC request and returns a string
+ * that will be sent back to the requester.
+ *
+ * Throwing an [RpcError] will send the error back to the requester.
+ *
+ * @see [LocalParticipant.registerRpcMethod]
+ */
+typealias RpcHandler = suspend (requestId: String, callerIdentity: Participant.Identity, payload: String, responseTimeoutMs: Int) -> String
```
bcherry, line 1620: the arguments should be provided as an `RpcInvocationData` data class so that we can add additional arguments in the future without breaking existing handlers

livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt
```diff
@@ -763,6 +776,369 @@ internal constructor(
...
+     * @param payload The payload to pass to the method.
+     * @param responseTimeoutMs Timeout for receiving a response after initial connection (milliseconds). Defaults to 10000.
+     * @return The response payload.
+     * @throws RpcError on failure. Details in [RpcError.message].
+     */
+    suspend fun performRpc(
+        destinationIdentity: Identity,
+        method: String,
+        payload: String,
+        responseTimeoutMs: Int = 10000,
```
bcherry, line 881: The duration parameters in the RPC feature should be more platform-native. I think in this case you should either use the kotlin native [Duration type](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.time/-duration/) type (introduced kotlin 1.6) or you should provide these as `Long` with the `Millis` suffix (e.g. `responseTimeoutMillis: Long = 10000`)

bcherry (changes requested):
looks good, just a couple notes about the interface!

bcherry (approved):

PR #579, opened 2025-01-09
Audio changes

davidliu:


livekit-android-sdk/src/main/java/io/livekit/android/audio/AudioSwitchHandler.kt
```diff
@@ -222,8 +223,8 @@ constructor(private val context: Context) : AudioHandler {
             listOf(
                 AudioDevice.BluetoothHeadset::class.java,
                 AudioDevice.WiredHeadset::class.java,
-                AudioDevice.Earpiece::class.java,
                 AudioDevice.Speakerphone::class.java,
+                AudioDevice.Earpiece::class.java,
```
tuan-jason, line 227: @davidliu I guess this fix is for: https://github.com/livekit/client-sdk-android/issues/580?
If that's the case, I'm curious what will become the active audio output in case we setup custom `preferredDeviceList` in which `Earpiece` is the first one, then disconnect/reconnect to room? Will that be `Speakerphone` or `Earpiece` with this fix?

davidliu: Didnt see that bug before making this change; our other SDKs use this ordering, so this is to align with that.

tuan-jason: I see. If you have time, could you take a look at the bug please? I think that might be related.
Thanks a lot.

adrian-niculescu:
Hi, @davidliu, the [doc comment fro `preferredDeviceList`](https://github.com/livekit/client-sdk-android/blob/e46da870e151955822d04a194a4dacedd823c00a/livekit-android-sdk/src/main/java/io/livekit/android/audio/AudioSwitchHandler.kt#L65) still has the old ordering with Earpiece being first before Speakerphone.

davidliu:
@adrian-niculescu Thanks for reminding me. Fixed in #584.

PR #619, opened 2025-03-04
expose RtpParameters.Encoding's minBitrateBps to VideoEncoding

andyli:
TLDR: Allow users to set a min bitrate. It is useful in situations where there is some minimal network bandwidth guarantee.

I observed that the video stream from my Android app, built with LiveKit Android, always starts with a lower bitrate (as low as ~100kbps), even if I've configured it to have a maximum bitrate of 8000kbps. After around 2 minutes, the bitrate goes up to 3000kbps and reaches almost 8000kbps after a while.

Setting the min bitrate to 6000kbps allowed me to get an initial bitrate of ~4000kbps from the beginning.


livekit-android-sdk/src/main/java/io/livekit/android/room/track/LocalVideoTrackOptions.kt
```diff
@@ -54,6 +54,7 @@ constructor(
 
 data class VideoEncoding(
     val maxBitrate: Int,
+    val minBitrate: Int,
```
davidliu, line 57: Can you make this an optional and set the default value to null instead? I'm not sure whether the underlying libwebrtc uses a default of 0.

andyli:
Rebased and minBitrate defaults to null now.

PR #626, opened 2025-03-19
Unpublish the screen sharing track on stop and introduce ScreenCaptureParams

jeankruger:
This PR unpublishes the screen sharing track when the user stops screen sharing using the system status button. Furthermore, it introduces a new ScreenCaptureParams model to allow customization of the foreground service notification and to listen for the onStop event


davidliu (approved):
LGTM!

PR #706, opened 2025-06-23
Fix RTCEngine not being set to connecting state

davidliu:


adrian-niculescu:
Thank you for this fix! Was replicating this in version `2.17.1`, but I wasn't sure if I needed to update the server, too, which was not possible for me at the time. Tested again now in `2.18.1` and it works 👍

PR #769, opened 2025-10-02
feat: TokenSource implementation

davidliu:


livekit-android-sdk/src/main/java/io/livekit/android/token/TokenSource.kt
```diff
@@ -0,0 +1,107 @@
...
+@SuppressLint("UnsafeOptInUsageError")
+@Serializable
+data class RoomAgentDispatch(
+    val agentName: String,
+    val metadata: String,
+)
+
+@SuppressLint("UnsafeOptInUsageError")
+@Serializable
+data class TokenSourceResponse(val serverUrl: String, val participantToken: String)
```
pblazej, line 58: Nit: `room_name` and `participant_name` can be added.

PR #807, opened 2025-11-18
Fix LocalParticipant jobs map leak on unpublish

adrian-niculescu:
- Ensure `LocalParticipant` stores jobs in a type-safe map keyed by LocalTrackPublication.
- Fix `unpublishTrack` to remove jobs using the publication key after cancelling the job.
- Prevent accumulation of cancelled jobs and associated publication references over the room lifetime.

CC @davidliu

### DIFF
--- a/livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt
+++ b/livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt
@@ -129,7 +129,7 @@ internal constructor(
             .mapNotNull { it as? LocalTrackPublication }
             .toList()
 
-    private val jobs = mutableMapOf<Any, Job>()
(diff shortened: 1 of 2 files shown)


livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt
```diff
@@ -129,7 +129,7 @@ internal constructor(
             .mapNotNull { it as? LocalTrackPublication }
             .toList()
 
-    private val jobs = mutableMapOf<Any, Job>()
+    private val jobs = mutableMapOf<LocalTrackPublication, Job>()
```
adrian-niculescu: @davidliu, I realize you intentionally keyed it `Any` to have it for wider future use. However, this hasn't applied in 8 months and you can always change it back to `Any` when you'll actually need it.

Issue #808, opened 2025-11-19
Release builds that use v2.23.0 crash at room.connect()


adrian-niculescu:
**Describe the bug**
Release builds that use v2.23.0 crash at room.connect().

**To Reproduce**
Have `minifyEnabled` and  `shrinkResources` enabled in the Android app's `build.gradle` like:

```groovy
 buildTypes {
...
        release {
           ...
            minifyEnabled = true
            shrinkResources = true
            proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
            ndk {
                debugSymbolLevel = 'FULL'
            }
        }
```

Root Cause
Commit https://github.com/livekit/client-sdk-android/commit/4bf585ecdd227a95fbeb2def10f9542e925389c9 introduced two data classes (`AgentAttributes` and `TranscriptionAttributes`) that use Klaxon for JSON parsing via reflection. However, no ProGuard rules were added to protect these classes from obfuscation.

Why It Crashes
When minifyEnabled = true and shrinkResources = true, ProGuard/R8 obfuscates class names and constructors
During r`oom.connect()`, the SDK calls `AgentAttributes.fromMap()`
Klaxon uses reflection to instantiate the obfuscated class
Klaxon can't find the constructor → crash: Couldn't find a suitable constructor for class b

PR #809, opened 2025-11-19
#808 Fixed ProGuard crash with Klaxon reflection-based classes

adrian-niculescu:
@davidliu , this fixes the release crash described in https://github.com/livekit/client-sdk-android/issues/808

Commit https://github.com/livekit/client-sdk-android/commit/4bf585ecdd227a95fbeb2def10f9542e925389c9 introduced two data classes (`AgentAttributes` and `TranscriptionAttributes`) that use Klaxon for JSON parsing via reflection. However, no ProGuard rules were added to protect these classes from obfuscation.

### DIFF
--- a/.changeset/silent-klaxon-proguard.md
+++ b/.changeset/silent-klaxon-proguard.md
@@ -1,0 +1,5 @@
+---
+"client-sdk-android": patch
+---
+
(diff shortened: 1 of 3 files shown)


davidliu (approved):
LGTM!

PR #819, opened 2025-11-24
Fix thread visibility issues in SignalClient

adrian-niculescu:
Added `@Volatile` to `isConnected`, `currentWs`, and `isReconnecting` fields in `SignalClient` to ensure proper memory visibility across threads.

These fields are accessed from multiple threads (OkHttp WebSocket callbacks and IO dispatcher coroutines) without synchronization, which could cause:
- Messages silently dropped if the callback thread doesn't see the updated `currentWs`
- Message sending failures if the IO dispatcher doesn't see the updated connection state

### DIFF
--- a/.changeset/fix-signal-client-volatile.md
+++ b/.changeset/fix-signal-client-volatile.md
@@ -1,0 +1,5 @@
+---
+"client-sdk-android": patch
+---
+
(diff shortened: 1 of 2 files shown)


davidliu (approved):
LGTM!

PR #849, opened 2026-01-14
perf: skip Klaxon parsing for empty agent attribute maps

YashJainSC:
## Problem

`AgentAttributes.fromStringMap()` was very slow on very first invocation for empty maps taking upto 1 seconds in some cases.

## Solution
Add early return when the input map is empty, returning a default 
`AgentAttributes()` directly. This avoids triggering Klaxon parsing 
when there's nothing to parse.

### DIFF
--- a/.changeset/sixty-radios-switch.md
+++ b/.changeset/sixty-radios-switch.md
@@ -1,0 +1,5 @@
+---
+"client-sdk-android": patch
+---
+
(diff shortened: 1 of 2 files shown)


davidliu:
LGTM! I think I'll also switch to kotlinx serialization here asap. Didn't think the first start time for klaxon was this drastic.

PR #902, opened 2026-03-26
fix IllegalArgumentException when rpc is over 15K

xianshijing-lk:
The current code will thro a IllegalArgumentException when RPC is over 15K
While the function comment says that it will return failure containing the exception


davidliu (approved):

xianshijing-lk:
Thanks @davidliu

PR #920, opened 2026-04-30
Fixed RTCEngine.addTrack leaking pendingTrackResolvers on cancellation

adrian-niculescu:
## Summary

`RTCEngine.addTrack` registers the publish continuation in `pendingTrackResolvers[cid]` but never cleans it up if the 20s `withDeadline` fires or the calling coroutine is cancelled — the `suspendCancellableCoroutine` block has no `invokeOnCancellation` handler. As a result, retrying the same `LocalAudioTrack` / `LocalVideoTrack` (same `MediaStreamTrack.id()` → same `cid`) hits the duplicate guard at `RTCEngine.kt:395` and throws `DuplicateTrackException`, blocking republish until any of:
- the server eventually responds for that `cid` (clears the entry via `onLocalTrackPublished`),
- the signal channel reconnects (`onClose` → `abortPendingPublishTracks`), or
- the engine closes.

For lost AddTrack requests or dropped responses on a healthy signal, the mic/camera publish stays poisoned indefinitely.

## Fix

Register a `cont.invokeOnCancellation` inside the same `synchronized(pendingTrackResolvers)` block that does the insert. The handler removes the entry only if it still references the same continuation (`pendingTrackResolvers[cid] === cont`) so a stale handler from a cancelled publish cannot evict a freshly-registered resolver bound to the same `cid` — necessary because the duplicate guard would still let the retry register before the previous handler runs in concurrent scenarios.

## Test plan

- [x] Added `addTrackTimeoutDoesNotPoisonRetry` and `addTrackCallerCancellationDoesNotPoisonRetry` in `RTCEngineMockE2ETest`. Both unregister the mock auto-responder so the first publish actually times out / stays in flight, then assert that a same-`cid` retry resolves cleanly when the server eventually replies.
- [x] Verified both tests **fail** without the fix with `Track with same ID local_cid has already been published!` — confirming they catch the original bug.
- [x] Both tests pass with the fix.
- [x] Full `livekit-android-test:testDebugUnitTest` suite passes (242 tests).
- [x] `./gradlew spotlessApply` clean.


davidliu (approved):
LGTM! Thanks for the PR!

PR #934, opened 2026-05-08
fix: prevent reconnect from hanging when server sends LEAVE or never responds

YashJainSC:
## Summary
- Resume `joinContinuation` when a LEAVE is received as the initial response to a reconnect (e.g. `STATE_MISMATCH`), so the reconnect loop can advance instead of hanging forever.
- Add a 10s timeout on the signal connect handshake. We noticed cases where the client connects the websocket but the request never reaches the server, leaving `client.reconnect()` suspended indefinitely.
- Add tests covering both clean WS close and failure close paths.

## Reproduction
- Reproduced the issue that reconnection loop hangs by using FORCE_SOFT_RECONNECT reconnect type. After disconnect once server removes the participant from room and we initiate reconnection with soft reconnect , server always sends leave with state mismatch error. Confirmed that it hangs after 1 attempt. With the fix now it doesn't hang and do full reconnect and succeed.

## Test plan
- [x] `reconnectDoesNotHangOnLeaveStateMismatchWithCleanClose` — server sends LEAVE then closes WS cleanly (`onClosed` path)
- [x] `reconnectDoesNotHangOnLeaveStateMismatchWithFailure` — server sends LEAVE then drops connection (`onFailure` path)
- [x] Existing tests pass

### DIFF
--- a/.changeset/neat-rabbits-draw.md
+++ b/.changeset/neat-rabbits-draw.md
@@ -1,0 +1,5 @@
+---
+"client-sdk-android": patch
+---
+
+fix: resume joinContinuation when LEAVE received during reconnect handshake to avoid reconnection loop hanging issue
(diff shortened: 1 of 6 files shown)


davidliu (approved):
Thanks for the PR!

Issue #602, opened 2025-02-19
Problem with bluetooth audio output


jgnt32:
I have issue with bluetooth audio. Sometimes sound coming from earspeaker (or Speakerphone) but not from connected headset. Event when livekit detect it sucesfully.

I have this AudioOptions

```
fun getAudioOptions() = AudioOptions(
            audioHandler = AudioSwitchHandler(context).apply {
                loggingEnabled = true
                manageAudioFocus = false
                preferredDeviceList = listOf(
                        AudioDevice.BluetoothHeadset::class.java,
                        AudioDevice.WiredHeadset::class.java,
                        AudioDevice.Earpiece::class.java,
                        AudioDevice.Speakerphone::class.java,
                )
            }
    )
```

And this what i see in logs 

> AS/AudioSwitch onDeviceConnected(BluetoothHeadset(name=LE_WH-1000XM3))
> AS/AudioSwitch Current user selected AudioDevice = null
> AS/AudioSwitch onActivate(BluetoothHeadset(name=LE_WH-1000XM3))
> AS/AudioSwitch onDeviceConnected(BluetoothHeadset(name=LE_WH-1000XM3))

AudioDeviceChangeListener also notice me whaе current selected audio device is bluetooth headset.

I tried to detect headset connecting on app side and setting it via **audioHandler.selectDevice(device)**
> AS/AudioSwitch Current user selected AudioDevice = BluetoothHeadset(name=LE_WH-1000XM3)

But result is same

I tried to debug sources and what i found:
Class AudioDeviceManager calling android.media.AudioManager.startBluetoothSco method and its failing in try catch with RemoteException.

```
@Deprecated 
public void startBluetoothSco() {
    final IAudioService service = getService();
    try {
        service.startBluetoothSco(mICallBack,
                getContext().getApplicationInfo().targetSdkVersion);
    } catch (RemoteException e) {
        throw e.rethrowFromSystemServer();
    }
}
```
Maybe its happening because this method is deprecated.

I tested it on Pixel 8 with android 15.

davidliu:
Can you provide the logs that happen when it errors?

jgnt32:
Unfortunately, I can't determine which logs belong to this error. The most similar thing is this short line

> 2025-02-19 12:14:08.201 12880-12912 BluetoothRemoteDevices  pid-12880                            E  RemoteException in calling IBluetoothConnectionCallback

I cant see whole stack trace in debug(

[...]

PR #941, opened 2026-05-15
docs(audio): clarify selectDevice() overrides preferredDeviceList; add headset-aware sample

daxiondi:
## Summary

`AudioSwitchHandler.selectDevice()` currently has a one-line KDoc:

> Select a specific audio device.

In practice this omission has caused several integrations to ship apps where audio is silently routed to the speakerphone even when the user has a wired or Bluetooth headset connected. The typical anti-pattern is implementing a "speakerphone on/off" toggle by unconditionally calling `selectDevice(Speakerphone::class.java)` (or `Earpiece`) at call start — which overrides the auto-selection driven by `preferredDeviceList` (`BluetoothHeadset > WiredHeadset > Speakerphone > Earpiece` by default) for the rest of the call, including across hot-plug events.

This is documentation-only. It:

1. Documents that `selectDevice()` is **sticky** and overrides `preferredDeviceList`'s automatic selection (and survives hot-plug).
2. Calls out the unconditional-`selectDevice(Speakerphone)` anti-pattern.
3. Adds a recommended pattern showing how to implement a speakerphone toggle that still respects headsets, by combining `selectDevice()` with `audioDeviceChangeListener` and a `preferSpeakerWhenNoHeadset` flag.
4. Reminds readers that `BLUETOOTH_CONNECT` (Android 12+, API 31) must be granted before `availableAudioDevices` will report any Bluetooth headset — a frequent source of "headset is connected in OS but app routes to speaker" bug reports.

Related community discussion: #602.

## Test plan

- [x] No code changes — only KDoc on `AudioSwitchHandler.selectDevice()`.
- [x] `./gradlew :livekit-android-sdk:spotlessKotlinCheck` passes.
- [x] The recommended pattern was verified end-to-end on a Pixel 7a (Android 14) with both wired and Bluetooth (A2DP/SCO) headsets — the headset is selected automatically and respected across plug/unplug events.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/livekit-android-sdk/src/main/java/io/livekit/android/audio/AudioSwitchHandler.kt
+++ b/livekit-android-sdk/src/main/java/io/livekit/android/audio/AudioSwitchHandler.kt
@@ -286,51 +286,20 @@ constructor(private val context: Context) : AudioHandler {
 
     /**
      * Select a specific audio device, overriding the automatic selection driven by
-     * [preferredDeviceList].
+     * [preferredDeviceList]. The selection is sticky: the chosen device stays selected
+     * whenever it is available, even as other devices are connected or disconnected,
+     * and AudioSwitch will resume it automatically if it temporarily disappears and
+     * later reconnects.
      *
-     * The selection is sticky: once you call this, the chosen device stays active even
-     * when the set of [availableAudioDevices] changes (for example, plugging or unplugging
-     * a wired headset, or connecting a Bluetooth headset). If you want to keep following
-     * the priority defined by [preferredDeviceList] across such hot-plug events, do not
-     * call [selectDevice] — leave the automatic selection in place.
+     * If your goal is only to change the priority order — for example, prefer
+     * Speakerphone over Earpiece as the fallback when no headset is present —
+     * set [preferredDeviceList] instead. Calling [selectDevice] is not required
+     * for that. In particular, calling `selectDevice(Speakerphone)` at the start
+     * of every call will silently override any wired or Bluetooth headset the
+     * user has connected.
      *
-     * A common pitfall is to unconditionally call `selectDevice(Speakerphone)` (or
-     * `Earpiece`) at the start of a call to implement a "speakerphone on/off" toggle.
-     * On devices where the user has a wired or Bluetooth headset connected, this will
-     * silently override the headset and route audio to the speaker/earpiece instead.
-     *
-     * Recommended pattern for a "speakerphone toggle" while still respecting headsets:
-     *
-     * ```kotlin
-     * // Treat the user toggle only as a fallback when no headset is present.
-     * private var preferSpeakerWhenNoHeadset: Boolean = true
-     *
-     * fun onUserToggleSpeakerphone(on: Boolean) {
-     *     preferSpeakerWhenNoHeadset = on
-     *     selectBestAudioDevice()
-     * }
-     *
-     * private fun installAudioRouting(handler: AudioSwitchHandler) {
-     *     handler.audioDeviceChangeListener = { _, _ -> selectBestAudioDevice() }
-     *     selectBestAudioDevice()
-     * }
-     *
-     * private fun selectBestAudioDevice() {
-     *     val handler = audioHandler ?: return
-     *     val devices = handler.availableAudioDevices
-     *     val target = devices.firstOrNull { it is AudioDevice.BluetoothHeadset }
-     *         ?: devices.firstOrNull { it is AudioDevice.WiredHeadset }
-     *         ?: if (preferSpeakerWhenNoHeadset)
-     *             devices.firstOrNull { it is AudioDevice.Speakerphone }
-     *         else
-     *             devices.firstOrNull { it is AudioDevice.Earpiece }
-     *     target?.let { handler.selectDevice(it) }
-     * }
-     * ```
-     *
-     * Note: on Android 12 (API 31) and above, the `BLUETOOTH_CONNECT` runtime
-     * permission must be granted before [availableAudioDevices] will report any
-     * Bluetooth headset, even if the OS shows the headset as connected.
+     * To return to fully automatic selection after a previous sticky call, pass
+     * `null`: `selectDevice(null)`.
      */
     @Synchronized
     fun selectDevice(audioDevice: AudioDevice?) {
--- a/livekit-android-sdk/src/main/java/io/livekit/android/audio/AudioSwitchHandler.kt
+++ b/livekit-android-sdk/src/main/java/io/livekit/android/audio/AudioSwitchHandler.kt
@@ -285,7 +285,52 @@ constructor(private val context: Context) : AudioHandler {
         get() = audioSwitch?.availableAudioDevices ?: listOf()
 
     /**
-     * Select a specific audio device.
+     * Select a specific audio device, overriding the automatic selection driven by
+     * [preferredDeviceList].
+     *
+     * The selection is sticky: once you call this, the chosen device stays active even
+     * when the set of [availableAudioDevices] changes (for example, plugging or unplugging
+     * a wired headset, or connecting a Bluetooth headset). If you want to keep following
+     * the priority defined by [preferredDeviceList] across such hot-plug events, do not
+     * call [selectDevice] — leave the automatic selection in place.
+     *
+     * A common pitfall is to unconditionally call `selectDevice(Speakerphone)` (or
+     * `Earpiece`) at the start of a call to implement a "speakerphone on/off" toggle.
+     * On devices where the user has a wired or Bluetooth headset connected, this will
+     * silently override the headset and route audio to the speaker/earpiece instead.
+     *
+     * Recommended pattern for a "speakerphone toggle" while still respecting headsets:
+     *
+     * ```kotlin
+     * // Treat the user toggle only as a fallback when no headset is present.
+     * private var preferSpeakerWhenNoHeadset: Boolean = true
+     *
+     * fun onUserToggleSpeakerphone(on: Boolean) {
+     *     preferSpeakerWhenNoHeadset = on
+     *     selectBestAudioDevice()
+     * }
+     *
+     * private fun installAudioRouting(handler: AudioSwitchHandler) {
+     *     handler.audioDeviceChangeListener = { _, _ -> selectBestAudioDevice() }
+     *     selectBestAudioDevice()
+     * }
+     *
+     * private fun selectBestAudioDevice() {
+     *     val handler = audioHandler ?: return
+     *     val devices = handler.availableAudioDevices
+     *     val target = devices.firstOrNull { it is AudioDevice.BluetoothHeadset }
+     *         ?: devices.firstOrNull { it is AudioDevice.WiredHeadset }
+     *         ?: if (preferSpeakerWhenNoHeadset)
+     *             devices.firstOrNull { it is AudioDevice.Speakerphone }
+     *         else
+     *             devices.firstOrNull { it is AudioDevice.Earpiece }
+     *     target?.let { handler.selectDevice(it) }
+     * }
+     * ```
+     *
+     * Note: on Android 12 (API 31) and above, the `BLUETOOTH_CONNECT` runtime
+     * permission must be granted before [availableAudioDevices] will report any
+     * Bluetooth headset, even if the OS shows the headset as connected.
      */
     @Synchronized
     fun selectDevice(audioDevice: AudioDevice?) {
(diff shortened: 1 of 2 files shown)


livekit-android-sdk/src/main/java/io/livekit/android/audio/AudioSwitchHandler.kt
```diff
@@ -285,7 +285,52 @@ constructor(private val context: Context) : AudioHandler {
         get() = audioSwitch?.availableAudioDevices ?: listOf()
 
     /**
-     * Select a specific audio device.
+     * Select a specific audio device, overriding the automatic selection driven by
```
adrian-niculescu: The recommended sample reaches for `audioDeviceChangeListener =`, but that setter is `@Deprecated("Use registerAudioDeviceChangeListener.")` (see `AudioSwitchHandler.kt:59-60`). Since this snippet is the public-facing pattern we're now shipping in KDoc, it should use the non-deprecated API:

```kotlin
private fun installAudioRouting(handler: AudioSwitchHandler) {
    handler.registerAudioDeviceChangeListener { _, _ -> selectBestAudioDevice() }
    selectBestAudioDevice()
}
```

Two reasons this matters beyond "it's deprecated":

1. The deprecated property is a single-listener slot — direct assignment silently clobbers any listener set by other integration code (or a future SDK-internal one). `registerAudioDeviceChangeListener` is additive, which is the whole reason it was introduced.
2. We're telling new integrators "here is the right way to do this." If the right way is a deprecated API, the next reader either copies it forward or wastes a cycle figuring out which one to trust.

While we're in this KDoc, a few smaller things are worth tightening so a copy-paste doesn't backfire:

- `selectBestAudioDevice()` calls `selectDevice()` from inside the device-change listener, and `selectDevice()` posts to the audio-switch handler thread. AudioSwitch already suppresses same-device redispatch, but a `target == handler.selectedAudioDevice` short-circuit makes the idempotence explicit and avoids needless no-op posts:

  ```kotlin
  val target = /* ... */
  if (target != null && target != handler.selectedAudioDevice) {
      handler.selectDevice(target)
  }
  ```

- `preferSpeakerWhenNoHeadset` is written from the UI thread (`onUserToggleSpeakerphone`) and read from the audio-switch handler thread (the listener). For a doc snippet this should be `@Volatile` or an `AtomicBoolean`, otherwise the visibility is unspecified.

- The new text says the selected device "stays active even when the set of availableAudioDevices changes." I checked the pinned `davidliu:audioswitch` bytecode and the actual mechanism is narrower than "the user selection is cleared on disconnect": `userSelectedAudioDevice` is **not** cleared when the selected device is unplugged. AudioSwitch keeps it set, temporarily routes to the best available device from `preferredDeviceList` while it's gone, and **resumes the selected device automatically if it reconnects**. The field is only auto-nulled in one special case (a `WiredHeadset` connects while `userSelectedAudioDevice` is an `Earpiece`). The sticky selection is otherwise only abandoned by calling `selectDevice(null)` (or `selectDevice` with a different device). The KDoc should describe this resume-on-reconnect behavior rather than implying the selection is dropped on disconnect.

- The draft says "do not call `selectDevice`" to stay on automatic selection but never says how to *undo* a prior sticky call. `selectDevice()` writes its argument straight into `userSelectedAudioDevice`, so `selectDevice(null)` is the supported way to return to fully automatic `preferredDeviceList` selection. Worth stating explicitly.

Optional but worth considering: for callers whose only goal is "speakerphone instead of earpiece as the fallback when no headset is present," they can use the default ordering or configure `preferredDeviceList` once before `start()` if they need a different static order. The dynamic listener is only required when the user's speaker toggle must change routing mid-call. Mentioning this above the snippet would steer most readers away from `selectDevice()` entirely, which is the KDoc's actual goal.

Suggested replacement:

````kotlin
/**
 * Select a specific audio device, overriding the automatic selection driven by
 * [preferredDeviceList].
 *
 * The selection is sticky: once set, that device stays selected whenever it is
 * available, even as other devices are connected or disconnected. If the selected
 * device becomes unavailable (for example, unplugged), AudioSwitch temporarily
 * routes to the best available device from [preferredDeviceList] and then
 * resumes the selected device automatically if it reconnects.
 *
 * If you want to keep following the priority defined by [preferredDeviceList],
 * do not call [selectDevice]. To undo a previous sticky selection and return to
 * fully automatic selection, call `selectDevice(null)`.
 *
 * A common pitfall is to unconditionally call `selectDevice(Speakerphone)` (or
 * `Earpiece`) at the start of a call to implement a "speakerphone on/off" toggle.
 * On devices where the user has a wired or Bluetooth headset connected, this will
 * override the headset and route audio to the speaker/earpiece instead.
 *
 * If the desired behavior is simply "prefer speakerphone over earpiece when no
 * headset is present", use the default device order or configure [preferredDeviceList]
 * before [start]. Use a listener only when the speakerphone toggle must change
 * routing while a call is already active.
 *
 * Example speakerphone toggle that still respects wired/Bluetooth headsets:
 *
 * ```kotlin
 * // Treat the user toggle only as a fallback when no headset is present.
 * @Volatile
 * private var preferSpeakerWhenNoHeadset: Boolean = true
 *
 * fun onUserToggleSpeakerphone(on: Boolean) {
 *     preferSpeakerWhenNoHeadset = on
 *     selectBestAudioDevice()
 * }
 *
 * private fun installAudioRouting(handler: AudioSwitchHandler) {
 *     handler.registerAudioDeviceChangeListener { _, _ -> selectBestAudioDevice() }
 *     selectBestAudioDevice()
 * }
 *
 * private fun selectBestAudioDevice() {
 *     val handler = audioHandler ?: return
 *     val devices = handler.availableAudioDevices
 *     val target = devices.firstOrNull { it is AudioDevice.BluetoothHeadset }
 *         ?: devices.firstOrNull { it is AudioDevice.WiredHeadset }
 *         ?: if (preferSpeakerWhenNoHeadset)
 *             devices.firstOrNull { it is AudioDevice.Speakerphone }
 *         else
 *             devices.firstOrNull { it is AudioDevice.Earpiece }
 *
 *     if (target != null && target != handler.selectedAudioDevice) {
 *         handler.selectDevice(target)
 *     }
 * }
 * ```
 *
 * Note: on Android 12 (API 31) and above, the `BLUETOOTH_CONNECT` runtime
 * permission must be granted before [availableAudioDevices] will report any
 * Bluetooth headset, even if the OS shows the headset as connected.
 */
````

**Review by adrian-niculescu — CHANGES_REQUESTED** [review]

File: livekit-android-sdk/src/main/java/io/livekit/android/audio/AudioSwitchHandler.kt [hunk 2]
```diff
@@ -285,7 +285,52 @@ constructor(private val context: Context) : AudioHandler {
...
+     *         ?: devices.firstOrNull { it is AudioDevice.WiredHeadset }
+     *         ?: if (preferSpeakerWhenNoHeadset)
+     *             devices.firstOrNull { it is AudioDevice.Speakerphone }
+     *         else
+     *             devices.firstOrNull { it is AudioDevice.Earpiece }
+     *     target?.let { handler.selectDevice(it) }
+     * }
+     * ```
+     *
+     * Note: on Android 12 (API 31) and above, the `BLUETOOTH_CONNECT` runtime
```
davidliu: BLUETOOTH_CONNECT is not required in the first place.

davidliu: BLUETOOTH_CONNECT is not required.

livekit-android-sdk/src/main/java/io/livekit/android/audio/AudioSwitchHandler.kt
davidliu: This is pretty verbose and could really use a rewrite to slim it down.

livekit-android-sdk/src/main/java/io/livekit/android/audio/AudioSwitchHandler.kt
```diff
@@ -285,7 +285,52 @@ constructor(private val context: Context) : AudioHandler {
...
+     *     preferSpeakerWhenNoHeadset = on
+     *     selectBestAudioDevice()
+     * }
+     *
+     * private fun installAudioRouting(handler: AudioSwitchHandler) {
+     *     handler.audioDeviceChangeListener = { _, _ -> selectBestAudioDevice() }
+     *     selectBestAudioDevice()
+     * }
+     *
+     * private fun selectBestAudioDevice() {
```
davidliu: This basically makes the user reimplement the audio device handling of AudioSwitchHandler. Users can just set [preferredDeviceList] to change the ordering if they just need a priority list.

davidliu (changes requested):
This whole thing needs quite a bit of work and basically reads like a "don't use this function". I'll probably just toss up a PR to update the docs here.

daxiondi:
Thanks @davidliu and @adrian-niculescu for the thorough review — both threads pointed at real problems with the first draft. I pushed `c4dedb7` to simplify the KDoc significantly. Summary of what changed and the reasoning:

### What was wrong with v1

The draft was trying to do two things at once: (a) document `selectDevice()`, and (b) ship a recommended pattern for "speakerphone toggle that still respects headsets". @davidliu correctly pointed out that the pattern reimplements what `AudioSwitchHandler` already does internally — and that "set `preferredDeviceList` once" is the right answer for the common case. I ran an experiment to verify this end-to-end before rewriting.

### Experiment (Pixel 7a, Android 14)

I stripped our app's wrapper down so that the listener / `selectBestAudioDevice` path is **never** called and `setSpeakerphoneOn` is a pure no-op. Only LiveKit's default `preferredDeviceList` is left to do the work. Two scenarios, same wrapper-no-op build:

```
// no headset connected
EXPERIMENT setSpeakerphoneOn=true (NO-OP): avail=[Speakerphone, Earpiece] selected=Speakerphone

// Bluetooth headset connected
EXPERIMENT setSpeakerphoneOn=true (NO-OP): avail=[BluetoothHeadset, Speakerphone, Earpiece] selected=BluetoothHeadset
```

Hot-plug during an active call (disconnecting / reconnecting the Bluetooth headset mid-call) also routed correctly without any `audioDeviceChangeListener` callback firing on our side — i.e. AudioSwitch is handling it entirely on its own. That confirms @davidliu's line 318 comment: the sample was strictly making things worse for the common case.

### What v2 does

- **Drop the sample entirely.** The KDoc now just points at `preferredDeviceList` for the "I just want a different priority order" case, which covers everyone except callers who genuinely need to flip routing on a running call.
- **Keep the anti-pattern callout**, narrowed to the one sentence that actually matters: unconditionally calling `selectDevice(Speakerphone)` at call start silently overrides headsets.
- **Fix the sticky-semantics description** (per @adrian-niculescu): the selection isn't dropped when the device disappears — AudioSwitch keeps `userSelectedAudioDevice` set, temporarily routes to the best available device from `preferredDeviceList`, and resumes the selected device automatically if it reconnects.
- **Document `selectDevice(null)`** (per @adrian-niculescu) as the supported undo path back to fully automatic selection.
- **Drop the `BLUETOOTH_CONNECT` note** (per @davidliu). Same Pixel 7a build with `BLUETOOTH_CONNECT` not granted still enumerated `BluetoothHeadset` correctly, so the note as written was overstated. There is some OEM-specific behavior on stricter ROMs that I can't reproduce on the device in hand, so rather than keep a half-true note in the SDK docs I'm just removing it.

Resulting diff is now 13 +/44 −, all in `selectDevice()`'s KDoc. Happy to iterate further if you'd still prefer to write this yourself — just wanted to take one pass at addressing the concrete review points first.

daxiondi:
@davidliu pushed `565f587` to further address the "reads like 'don't use this function'" feedback. The previous revision still carried an anti-pattern callout that kept the overall tone negative; v3 rewrites the KDoc as a neutral how-to-use, two sentences total:

```kotlin
/**
 * Select a specific audio device. The selection is sticky: it overrides
 * [preferredDeviceList], persists across device hot-plug, and is restored
 * automatically if the device temporarily disappears and later reconnects.
 * Pass `null` to clear a sticky selection.
 *
 * If you only need to change which device is preferred when several are
 * available, set [preferredDeviceList] instead.
 */
```

Five lines of KDoc, no warnings. The substantive technical content from @adrian-niculescu (sticky/resume semantics + `selectDevice(null)` as the undo path) is preserved; the "set `preferredDeviceList` instead" recommendation you requested is preserved; the explicit anti-pattern warning is dropped because the existence of `preferredDeviceList` in the recommendation already implies it.

Would appreciate another look when you have a moment.

davidliu (approved):

davidliu (approved):

PR #948, opened 2026-05-22
Increase data packet size check to 64KB in raw data packets

1egoman:
Most other client sdks (I referenced `client-sdk-js`, `rust-sdks`, and `client-sdk-swift`) allow a user to send a raw data packet via `localParticipant.publishData` which can be as long the user wants.

`client-sdk-android` on the other hand looks to be special, and if you send longer than 15kb, it will bail early and throw an error:
```kotlin
if (data.size > RTCEngine.MAX_DATA_PACKET_SIZE) {
    return Result.failure(IllegalArgumentException("cannot publish data larger than " + RTCEngine.MAX_DATA_PACKET_SIZE))
}
```

My initial thinking is that it might make sense to remove this check on android because none of the other sdks do it this way, and this is a low level api where the assumption would be that a user knows what they are doing / would understand the potential implications of sending a very long data packet. It wouldn't be a breaking change because removing the check is "widening" the api interface / all existing callers would still work.

I've tested this locally by sending data packets larger than 15kb to the android example app and it is able to receive it properly. If there's any more extensive testing that needs to be done here, please let me know.

### DIFF
--- a/.changeset/tricky-carpets-fall.md
+++ b/.changeset/tricky-carpets-fall.md
@@ -1,0 +1,5 @@
+---
+"client-sdk-android": patch
+---
+
+Remove data packet size check in localParticipant.publishData to match other sdks
--- a/.changeset/tricky-carpets-fall.md
+++ b/.changeset/tricky-carpets-fall.md
@@ -2,4 +2,4 @@
 "client-sdk-android": patch
 ---
 
-Remove data packet size check in localParticipant.publishData to match other sdks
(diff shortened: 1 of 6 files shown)


davidliu:
From testing, sending packets over the `max-message-size` will cause the data channel to close, so instead increased the limit from 15KB to what the server negotiates (64KB - 1). Left a TODO since the limit is still hardcoded rather than examining what is negotiated.

davidliu (approved):

PR #961, opened 2026-06-08
Add support for generic request response

davidliu:


livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt
```diff
@@ -130,6 +138,8 @@ internal constructor(
 
     private val jobs = mutableMapOf<LocalTrackPublication, Job>()
 
+    private val pendingSignalRequests = Collections.synchronizedMap(mutableMapOf<Int, CompletableDeferred<Unit>>())
```
lukasIO, line 141: apart from the existing timeout, we should probably also cleanup + reject pending entries on room disconnects?

PR #971, opened 2026-06-24
fix(tracks): dispose video source and stop transceiver on unpublish

### LINKED ISSUE #521: Memory leaks produced by publishing/unpublishing camera tracks
**Describe the bug**
See https://github.com/livekit/client-sdk-swift/issues/420. I found that the same memory leak problem appears on Android for every publish -> unpublish track operation. 

**To Reproduce**
Publish then unpublish a camera track, repeat these steps multiple times and you will see that the memory increases. The memory will be deallocated when the room is disconnected.

**Expected behavior**
After every camera track unpublish the memory should decrease.

**Screenshots**
Android Memory Leaks

**Device Info:**
 - Device: Samsung Galaxy A33
 - OS: Android 13
 - LiveKit SDK version: v2.6.0

**Additional context**
See https://github.com/livekit/client-sdk-swift/issues/420#issuecomment-2399552933

adrian-niculescu:
Fixes #521.

Two native leaks build up on every video publish/unpublish cycle:

- `LocalVideoTrack.dispose()` released the track and capturer but not the `VideoSource`, leaking the source's native memory for the lifetime of the process. `restartTrack()` already disposes the old source, so terminal dispose was the only path missing it.
- `unpublishTrack()` removed the track from its sender but never stopped the `RtpTransceiver`. A new transceiver is created on every publish, so they accumulated until the connection closed. Stopping the now-unused video transceiver releases it and frees the SDP m-section for reuse. Same approach as the iOS fix (livekit/client-sdk-swift#420).

Tests: added coverage that dispose disposes the source and that unpublishing a video track stops its transceiver.

### DIFF
--- a/.changeset/video-track-publish-unpublish-leaks.md
+++ b/.changeset/video-track-publish-unpublish-leaks.md
@@ -1,0 +1,5 @@
+---
+"client-sdk-android": patch
+---
+
+Fix native memory leaks on video track publish/unpublish cycles (#521). `LocalVideoTrack.dispose()` now disposes its backing `VideoSource`, which was previously left undisposed and leaked for the lifetime of the process (only the track and capturer were released). Unpublishing a video track now also stops its `RtpTransceiver`, along with any extra transceivers added for backup codecs; since a new transceiver is created on every publish, removing the track from its sender alone left them retained until the connection closed.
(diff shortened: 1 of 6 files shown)


davidliu:
Looks good, I'll test it out.

adrian-niculescu:
> Looks good, I'll test it out.

Thanks! I've been using it on Android 13 and Android 17 (latest beta) since last week and haven't run into trouble.

davidliu:
Tested and leaks no longer happen with repeated publish/unpublish.

adrian-niculescu:
Thank you!

PR #973, opened 2026-06-29
fix: improve initial video quality by setting x-google-start-bitrate for all video codecs

xianshijing-lk:
## Problem

Published video is blurry for the first 5-15 seconds. WebRTC's bandwidth estimator starts near 300 kbps and ramps slowly, so the encoder spends the opening seconds far below the track's target bitrate.

## Solution

Seed the estimator with `x-google-start-bitrate` for **all** video codecs, not just SVC. The hint is 90% of the track's target bitrate (10% headroom for BWE), capped at **1 Mbps** for camera and skipped below a **300 kbps** target.

- **Why cap camera at 1 Mbps.** Seeding too high causes an overshoot the estimator must recover from, which showed up as higher first-frame latency and choppiness over the first ~30s on constrained links. 1 Mbps still carries 720p, and BWE ramps from there quickly. Screen shares are exempt: they are published at high bitrates for text legibility, where a conservative start is more costly than a brief overshoot.
- **Why a 300 kbps floor.** Below that, seeding above the real capacity costs more than the ramp it saves.
- **Known tradeoff.** High-bitrate SVC cameras previously started at 0.7x with no ceiling (a 3 Mbps VP9 publish started at 2100 kbps; it now starts at 1000). Above roughly 1.4 Mbps targets the new start is lower, which can lengthen the ramp on links that can carry the target — and is safer on links that can't. Same values as client-sdk-js. A follow-up will make the seed adaptive using previous BWE estimates rather than a fixed cap.

## One connection-level value, written once

These codec fmtp params are connection-scoped, not m-section-scoped. libwebrtc reads them per m-section (`WebRtcVideoSendChannel::ApplyChangedParams` → `GetBitrateConfigForCodec`) but pushes the result into the shared `Call` via `SetSdpBitrateParameters`, where `RtpBitrateConfigurator` holds **one** config for the entire peer connection. Two m-sections carrying different values is last-writer-wins, decided by SDP order.

So the SDK computes a single value — the max hint across video m-sections in the first offer containing local video — and writes that same value to every video m-section.

It is written **once** per publisher connection, because the value persists: `RtpBitrateConfigurator` retains `start_bitrate_bps` in its stored config and `RtpTransportControllerSend::OnNetworkRouteChanged` re-applies it from `GetConfig()` on every relevant route change. A WiFi-to-cellular handover therefore re-seeds the estimator from this hint with no renegotiation. Rewriting it on later offers is at best a no-op (libwebrtc ignores an unchanged value, and only re-reads it when the send codec changes) and at worst restarts a converged estimator. A full reconnect builds a new peer connection with a new estimator, and seeds it again.

## `x-google-max-bitrate` is no longer written

**This is a behavior change.** The same Call-level promotion turned a per-track cap into a ceiling on total send bandwidth for the whole connection — a default camera publish emitted `x-google-max-bitrate=2310`, which could starve a concurrent 3 Mbps screen share depending on m-section order. libwebrtc carries a TODO conceding this is wrong: *"codec max bitrate should probably not affect global call max bitrate."*

Per-track and per-layer limits are still enforced through `RtpParameters.Encoding.maxBitrateBps`, which is genuinely scoped per encoding. client-sdk-js and the Rust SDK never write the SDP value either, so this is cross-SDK parity rather than an Android-only change.

## Also

- `TrackBitrateInfo` / `TrackBitrateInfoKey` are now `internal`. Both were `@suppress`ed and only reachable via a `@VisibleForTesting` helper; they were public solely to be visible from the test module, which `friendPaths` has made unnecessary since #925. `TrackBitrateInfo.maxBitrate` is renamed to `targetBitrateKbps` — it always carried kbps, and nothing in the old name said so.

Degradation-preference defaults are **not** part of this PR; they shipped separately in #991.

## Test plan

- [x] `SdpMungingTest` covers the connection-level value, the 300 kbps floor, and the cap
- [x] Full unit suite green (331 tests)
- [ ] Verify video is sharp from the start when publishing, across VP8, VP9, H264, AV1
- [ ] Verify camera + screen share concurrently: neither throttles the other
- [ ] Verify BWE adapts when the network cannot carry the seeded start bitrate


protocol
adrian-niculescu, line 1: This appears to be unrelated to the bitrate/SDP change.

livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt
```diff
@@ -697,9 +698,9 @@ internal constructor(
             track.statsGetter = engine.createStatsGetter(transceiver.sender)
 
             val finalOptions = options
-            // Handle trackBitrates
+            // Handle trackBitrates - apply start bitrate for all video codecs to prevent initial blurriness
             if (encodings.isNotEmpty()) {
-                if (finalOptions is VideoTrackPublishOptions && isSVCCodec(finalOptions.videoCodec) && encodings.firstOrNull()?.maxBitrateBps != null) {
+                if (finalOptions is VideoTrackPublishOptions && isVideoCodec(finalOptions.videoCodec) && encodings.firstOrNull()?.maxBitrateBps != null) {
```
adrian-niculescu: Switching this gate from `isSVCCodec` to `isVideoCodec` pulls simulcast codecs (VP8 by default, plus H264) into this path, but `maxBitrate` is still taken from `encodings.first()`. For simulcast, `computeVideoEncodings` adds encodings smallest-to-largest (the `sortedByDescending { calculateScaleDown }` ordering), so `encodings.first()` is the lowest layer. A default 16:9 publish starts at H180 = 160 kbps. `registerTrackBitrateInfo` feeds that into `ensureCodecBitrates`, which writes a codec-level `x-google-start-bitrate`/`x-google-max-bitrate` from the low-layer value, capping the full-resolution layer far below its target. SVC kept a single full-bitrate encoding, which is why this was SVC-only before. For simulcast you'd need the top layer's bitrate (or to skip the codec-level cap).

livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt
```diff
@@ -1481,7 +1482,8 @@ data class VideoTrackPublishDefaults(
     override val videoCodec: String = VideoCodec.VP8.codecName,
     override val scalabilityMode: String? = null,
     override val backupCodec: BackupVideoCodec? = null,
-    override val degradationPreference: RtpParameters.DegradationPreference? = null,
+    // Default to MAINTAIN_RESOLUTION to prevent initial video blurriness
+    override val degradationPreference: RtpParameters.DegradationPreference? = RtpParameters.DegradationPreference.MAINTAIN_RESOLUTION,
```
adrian-niculescu: This flips the default for every video publisher, not just the slow-ramp case. `MAINTAIN_RESOLUTION` holds resolution by dropping framerate under congestion, which suits screen share but turns camera video into low/stuttering framerate on constrained networks; WebRTC's default for camera content is maintain-framerate. The same change is in `VideoTrackPublishOptions` below, and the KDoc on the abstract `degradationPreference` still reads "null value indicates default value (maintain framerate)", which now contradicts the default. Consider scoping `MAINTAIN_RESOLUTION` to screen share rather than making it the global default, and updating the KDoc.

munging.patch
```diff
@@ -0,0 +1,591 @@
+diff --git a/gradle/libs.versions.toml b/gradle/libs.versions.toml
```
adrian-niculescu: This 591-line patch file looks accidentally committed. It's a separate SDP-parser refactor (JainSDP `MediaDescription` to a hand-rolled `SdpMediaSection`), unrelated to the bitrate change, and shouldn't be in the tree. It also trips `git diff --check` on trailing whitespace. Please drop it before merge.

adrian-niculescu (changes requested):
The start-bitrate consolidation is a reasonable idea, but a few things need addressing before this lands. Inline notes cover the simulcast bitrate regression, the global degradation-preference default, and the stray patch file.

One more that can't be anchored to a changed line: `SdpMungingTest.ensureCodecBitratesTest` (livekit-android-test/src/test/java/io/livekit/android/room/SdpMungingTest.kt) still asserts `x-google-start-bitrate=700000` for a 1 Mbps target, but the multiplier is now 0.9, so the munge emits 900000. That test fails as-is and needs updating.

protocol
xianshijing-lk: Thanks, yes, it should removed, it got checked in by mistake

xianshijing-lk:
> The start-bitrate consolidation is a reasonable idea, but a few things need addressing before this lands. Inline notes cover the simulcast bitrate regression, the global degradation-preference default, and the stray patch file.
> 
> One more that can't be anchored to a changed line: `SdpMungingTest.ensureCodecBitratesTest` (livekit-android-test/src/test/java/io/livekit/android/room/SdpMungingTest.kt) still asserts `x-google-start-bitrate=700000` for a 1 Mbps target, but the multiplier is now 0.9, so the munge emits 900000. That test fails as-is and needs updating.

Thanks for pointing out, I am discussing with the team on the proper fix here. 
There are two alternatives here :
1. Only apply start bitrate munging for SVC codecs (VP9, AV1) - simpler, matches Android fix
 2. Use the sum of all encodings' bitrates : matches Rust SDK approach

livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt
xianshijing-lk: I ran a series of experiments comparing MAINTAIN_FRAMERATE and MAINTAIN_RESOLUTION under constrained uplink conditions using custom Network Link Conditioner profiles.

Network Profiles: 
Poor 3G Uplink
Uplink: 300 Kbps
Packet loss: 3%
Downlink: Unbounded
Poor LTE Uplink
Uplink: 3 Mbps
Packet loss: 1%
Downlink: Unbounded
Results
Poor 3G (540p capture)

Both degradation preferences were tested at a 540p capture resolution.

MAINTAIN_FRAMERATE
The first ~10s of video is blurry, and it improves its resolution after that.
MAINTAIN_RESOLUTION
Preserves image clarity by holding a higher resolution.

Don't see much difference on the smoothness of the videos.

Poor LTE (720p capture)

Both degradation preferences were tested at a 720p capture resolution.

The same observation above.

Poor 3G (1080p capture)

An additional experiment was run using a 1080p capture resolution over the 300 Kbps uplink profile.

I start to notice differences of framerate and resolution here. In MaintainFrame mode, the video stays at lower resolution and it does look smoother;
While in MaintainResolution mode, the video stays at high resolution, but less smooth.
I don't see serious choppiness from either modes though.


Poor 3G (both uplink and downlink)

In this experiment, both uplink and downlink were bandwidth constrained, which produced noticeable choppiness regardless of the degradation preference. With the downlink no longer being the bottleneck, the differences between the two modes became much clearer.

Overall, the severe choppiness observed previously appears to have been caused by the network being constrained in both directions rather than by the degradation preference itself.

Personal Impression

When bandwidth is insufficient, the trade off between image quality and motion smoothness is largely subjective. Personally, I slightly prefer maintaining resolution, as I find occasional choppiness less distracting than a combination of choppiness and significant blurriness.

We are still having some discussions internally to figure out the best default option here, and appreciate any input / opinions here.

adrian-niculescu: The experiments are useful, and MAINTAIN_FRAMERATE is the right baseline: with the preference unset, the WebRTC build this SDK ships (io.github.webrtc-sdk:android-prefixed 144.7559.05) resolves ordinary camera content to MAINTAIN_FRAMERATE ([`GetDegradationPreference()` in `webrtc_video_engine.cc`](https://github.com/webrtc-sdk/webrtc/blob/f47af7bc965851090bef9fce9a4284f468d20a44/media/engine/webrtc_video_engine.cc#L1930-L1963) on `m144_release`, BALANCED sits behind the WebRTC-Video-BalancedDegradation field trial, which this SDK doesn't set), and screenshare already resolves to MAINTAIN_RESOLUTION. Two things still give me pause about changing the default:

1. Network conditioning only exercises half of what this knob controls. Degradation preference also drives CPU overuse adaptation. Under MAINTAIN_FRAMERATE an overloaded encoder downscales; under MAINTAIN_RESOLUTION it drops frames instead. A low-end phone that can't encode 720p30 won't show up in conditioner runs on a dev device, and Android has a lot of those phones.

2. Your own results describe the bandwidth-constrained tradeoff as subjective and content dependent, and the start bitrate fix already handles the blurry first seconds on links that can carry the target. What's left is a steady state preference, which is a product-level default to pick deliberately and across the SDKs, not on Android alone inside a bitrate PR. The option is exposed per publish today, so apps that prefer resolution can already opt in.

I'd take the default change out of this PR and run the default discussion separately.

adrian-niculescu:
> > The start-bitrate consolidation is a reasonable idea, but a few things need addressing before this lands. Inline notes cover the simulcast bitrate regression, the global degradation-preference default, and the stray patch file.
> > One more that can't be anchored to a changed line: `SdpMungingTest.ensureCodecBitratesTest` (livekit-android-test/src/test/java/io/livekit/android/room/SdpMungingTest.kt) still asserts `x-google-start-bitrate=700000` for a 1 Mbps target, but the multiplier is now 0.9, so the munge emits 900000. That test fails as-is and needs updating.
> 
> Thanks for pointing out, I am discussing with the team on the proper fix here. There are two alternatives here :
> 
> 1. Only apply start bitrate munging for SVC codecs (VP9, AV1) - simpler, matches Android fix
> 2. Use the sum of all encodings' bitrates : matches Rust SDK approach

Glad I can help. I'd go with option 2. Option 1 is what Android and JS already do today, so the default VP8 simulcast publish would still ramp up slowly and the original problem stays unfixed. All that would be left of this PR is the multiplier bump and the degradation default.

Option 2 is also the correct one on the merits: `x-google-start-bitrate` seeds the send-side BWE for the whole connection, so it should reflect the total bitrate we intend to send. For simulcast that's the sum of the layers. For SVC the single encoding already is the total, so summing changes nothing there. And it's what the Rust SDK does today: `rtc_session.rs` [sums `max_bitrate` across encodings](https://github.com/livekit/rust-sdks/blob/95187dff9324e474ebb27a6ee7d2213f9a1caad3/livekit/src/rtc_engine/rtc_session.rs#L1815-L1819), `peer_transport.rs` [applies 0.9 to all video codecs](https://github.com/livekit/rust-sdks/blob/95187dff9324e474ebb27a6ee7d2213f9a1caad3/livekit/src/rtc_engine/peer_transport.rs#L181-L199).

Before the notes, a correction to my earlier inline comment. `computeVideoEncodings` returns the encodings largest first ([`encodings.reverse()` at the end](https://github.com/livekit/client-sdk-android/blob/bf010574279928cd3db82806218d1509667bf847/livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt#L883-L884)), so `encodings.first()` in the earlier revision was the top layer, not the lowest one as I wrote. The mismatch is milder than I described but still real: a default 720p simulcast publish is 160 + 450 + 1700 kbps, so the top layer alone understates the total by about 25%, and writing it into `x-google-max-bitrate` caps the whole video send below what the three layers need. The code comment added in the latest revision repeats my smallest-to-largest claim, so that needs fixing too.

A few notes for the Android version:

1. `ensureCodecBitrates` writes both `x-google-start-bitrate` and `x-google-max-bitrate` from the same `TrackBitrateInfo.maxBitrate`, so the registered value needs to be the sum (or just drop the max for non-SVC codecs). Any single layer's value understates the aggregate.
2. Copy the Rust guard that skips the hint when the total is under 300 kbps. Below that, the hint hurts more than it helps.
3. The 0.7 to 0.9 bump needs its own justification. Was 0.9 actually validated on constrained uplinks, in the Rust rollout or elsewhere, or did it just come over with the patch? Starting at 90% of target on a link that can't carry it means loss and a hard BWE backoff in the first seconds, which looks worse than the slow ramp we're trying to fix. If there's data behind 0.9, link it. If not, I'd keep 0.7 and bump it separately. Also note this changes the shipped SVC behavior from 0.7 to 0.9 as a side effect.
4. Keep the per-cid `trackBitrates` registration, don't copy the Rust structure. Rust keeps one value per peer connection, so the last published video track wins and every video m-section gets rewritten. The per-msid matching here gets it right per track when camera and screenshare are both up. The change is small anyway: drop the `isSVCCodec` gate on registration and sum `maxBitrateBps` over the encodings (still divided by 1000, these fmtp values are kbps).

On consistency: JS and Rust already disagree (SVC-only at 0.7 vs all codecs at 0.9), so that argument works for either option. Whatever you pick here is probably what the other SDKs should converge on, and Rust shipping the sum approach suggests that's the direction.

On the degradation preference default, I replied in the review thread with the experiments. The submodule bump note still applies either way.

livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt
xianshijing-lk: Agreed. I changed the degradationPreference to Maintain_Resolution for screenshare, and back to Maintain_Framerate for Camera.

livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt
xianshijing-lk: I changed the default degradation preference to Maintain_Resolution for screen sharing, Maintain_Framerate for camera sources, and Balanced for other sources (which aren't supported today).

This matches the behavior we're standardizing across the other LiveKit SDKs.

I'm happy to move these changes into a separate PR if you think they will need more discussions.

xianshijing-lk:
> > > The start-bitrate consolidation is a reasonable idea, but a few things need addressing before this lands. Inline notes cover the simulcast bitrate regression, the global degradation-preference default, and the stray patch file.
> > > One more that can't be anchored to a changed line: `SdpMungingTest.ensureCodecBitratesTest` (livekit-android-test/src/test/java/io/livekit/android/room/SdpMungingTest.kt) still asserts `x-google-start-bitrate=700000` for a 1 Mbps target, but the multiplier is now 0.9, so the munge emits 900000. That test fails as-is and needs updating.
> > 
> > 
> > Thanks for pointing out, I am discussing with the team on the proper fix here. There are two alternatives here :
> > 
> > 1. Only apply start bitrate munging for SVC codecs (VP9, AV1) - simpler, matches Android fix
> > 2. Use the sum of all encodings' bitrates : matches Rust SDK approach
> 
> Glad I can help. I'd go with option 2. Option 1 is what Android and JS already do today, so the default VP8 simulcast publish would still ramp up slowly and the original problem stays unfixed. All that would be left of this PR is the multiplier bump and the degradation default.
> 
> Option 2 is also the correct one on the merits: `x-google-start-bitrate` seeds the send-side BWE for the whole connection, so it should reflect the total bitrate we intend to send. For simulcast that's the sum of the layers. For SVC the single encoding already is the total, so summing changes nothing there. And it's what the Rust SDK does today: `rtc_session.rs` [sums `max_bitrate` across encodings](https://github.com/livekit/rust-sdks/blob/95187dff9324e474ebb27a6ee7d2213f9a1caad3/livekit/src/rtc_engine/rtc_session.rs#L1815-L1819), `peer_transport.rs` [applies 0.9 to all video codecs](https://github.com/livekit/rust-sdks/blob/95187dff9324e474ebb27a6ee7d2213f9a1caad3/livekit/src/rtc_engine/peer_transport.rs#L181-L199).
> 
> Before the notes, a correction to my earlier inline comment. `computeVideoEncodings` returns the encodings largest first ([`encodings.reverse()` at the end](https://github.com/livekit/client-sdk-android/blob/bf010574279928cd3db82806218d1509667bf847/livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt#L883-L884)), so `encodings.first()` in the earlier revision was the top layer, not the lowest one as I wrote. The mismatch is milder than I described but still real: a default 720p simulcast publish is 160 + 450 + 1700 kbps, so the top layer alone understates the total by about 25%, and writing it into `x-google-max-bitrate` caps the whole video send below what the three layers need. The code comment added in the latest revision repeats my smallest-to-largest claim, so that needs fixing too.
> 
> A few notes for the Android version:
> 


> 4. Keep the per-cid `trackBitrates` registration, don't copy the Rust structure. Rust keeps one value per peer connection, so the last published video track wins and every video m-section gets rewritten. The per-msid matching here gets it right per track when camera and screenshare are both up. The change is small anyway: drop the `isSVCCodec` gate on registration and sum `maxBitrateBps` over the encodings (still divided by 1000, these fmtp values are kbps).
> 
> On consistency: JS and Rust already disagree (SVC-only at 0.7 vs all codecs at 0.9), so that argument works for either option. Whatever you pick here is probably what the other SDKs should converge on, and Rust shipping the sum approach suggests that's the direction.
> 
> On the degradation preference default, I replied in the review thread with the experiments. The submodule bump note still applies either way.

Thanks for the technical insights. 

> 1. `ensureCodecBitrates` writes both `x-google-start-bitrate` and `x-google-max-bitrate` from the same `TrackBitrateInfo.maxBitrate`, so the registered value needs to be the sum (or just drop the max for non-SVC codecs). Any single layer's value understates the aggregate.

it is addressed:
  livekit-android-sdk/src/main/java/io/livekit/android/room/PeerConnectionTransport.kt
  - Line 497: newFmtpConfig = "$newFmtpConfig;x-google-start-bitrate=$startBitrate"
  - Line 500: newFmtpConfig = "$newFmtpConfig;x-google-max-bitrate=${trackBr.maxBitrate}"

  The startBitrate calculation (lines 484-488):
  val calculatedStartBitrate = (trackBr.maxBitrate * startBitrateMultiplier).roundToLong()
  val startBitrate = if (trackBr.isScreenShare) {
      calculatedStartBitrate  // No cap for screen share
  } else {
      minOf(calculatedStartBitrate, maxStartBitrateKbps)  // Capped at 1 Mbps
  }

  For camera (720p simulcast):
  - x-google-start-bitrate = 1000 kbps (capped)
  - x-google-max-bitrate = 2310 kbps (full sum: 160 + 450 + 1700)

  For screen share (e.g., 3000 kbps target):
  - x-google-start-bitrate = 2700 kbps (90%, no cap - text clarity needs high bitrate immediately)
  - x-google-max-bitrate = 3000 kbps (full value)

  The 1 Mbps cap only affects the start hint for camera tracks, not the max bitrate constraint.


> 2. Copy the Rust guard that skips the hint when the total is under 300 kbps. Below that, the hint hurts more than it helps.

Done with addressing this corner case by skipping the hint if target bitrate is below 300, though I don't think it will affect the performance as it will start ramping up from 0.9xtarget maxBitrate to maxBitrate, which should be very quickly. 


> 3. The 0.7 to 0.9 bump needs its own justification. Was 0.9 actually validated on constrained uplinks, in the Rust rollout or elsewhere, or did it just come over with the patch? Starting at 90% of target on a link that can't carry it means loss and a hard BWE backoff in the first seconds, which looks worse than the slow ramp we're trying to fix. If there's data behind 0.9, link it. If not, I'd keep 0.7 and bump it separately. Also note this changes the shipped SVC behavior from 0.7 to 0.9 as a side effect.

The min(1Mbps, 0.9 multiplier) is being applied consistently across all SDKs (Rust, JS, Android) as part of this alignment effort.
See JS PR: https://github.com/livekit/client-sdk-js/pull/1987/changes

We believe 0.9xtargetBitrate with the 1 Mbps cap provides reasonable bounds, and a reasonable UX. 

I'm planning to improve this further with smarter bitrate selection based on network conditions (e.g., using previous BWE estimates or connection quality signals, signaling latency, ..etc) as a follow up. The PR is WIP. 



> 4. Keep the per-cid `trackBitrates` registration, don't copy the Rust structure. Rust keeps one value per peer connection, so the last published video track wins and every video m-section gets rewritten. The per-msid matching here gets it right per track when camera and screenshare are both up. The change is small anyway: drop the `isSVCCodec` gate on registration and sum `maxBitrateBps` over the encodings (still divided by 1000, these fmtp values are kbps).
> 
> On consistency: JS and Rust already disagree (SVC-only at 0.7 vs all codecs at 0.9), so that argument works for either option. Whatever you pick here is probably what the other SDKs should converge on, and Rust shipping the sum approach suggests that's the direction.

These should all be addressed.

xianshijing-lk:
Hi @adrian-niculescu , could you please take another look ? thanks.

munging.patch
xianshijing-lk: this was moved, it was checked in mistakenly by claude.

protocol
```diff
@@ -1 +1 @@
-Subproject commit 8381f2180c45ab926b3ebf19df0608f1dadcac1e
+Subproject commit 4c05a3325ec35760bee1c0bfe57b7011604a124f
```
adrian-niculescu: This moves the `protocol` submodule pointer backwards, from `8381f21` (current `main`) to `4c05a33`, an ancestor ~303 commits / 8 months older. Looks accidental, probably a stale local submodule checkout that got staged with the fix. On merge it would roll the protobuf definitions back for everyone.

It is also not just a silent revert: the branch does not compile against `4c05a33`. A clean `:livekit-android-sdk:compileReleaseKotlin` fails with unresolved references to symbols added to the protocol after that commit (`AGENT_ERROR` in RoomEvent.kt, `PUBLISH_DATA_TRACK_RESPONSE`/`UNPUBLISH_DATA_TRACK_RESPONSE`/`DATA_TRACK_SUBSCRIBER_HANDLES` in SignalClient.kt, `clientProtocol`/`CONNECTOR`/`BRIDGE` in Participant.kt and ClientInfo.kt). CI currently stops earlier at the Spotless check, so this will surface as the next failure once that is fixed.

Restoring it against `main`:

```bash
git checkout main -- protocol && git commit -m "chore: restore protocol submodule pointer"
```

livekit-android-sdk/src/main/java/io/livekit/android/room/PeerConnectionTransport.kt
```diff
@@ -469,14 +480,29 @@ fun ensureCodecBitrates(
             ?: continue
         val codecPayload = rtp.payload
 
+        // Skip start bitrate hint for very low bitrate tracks - the hint hurts more than it helps
+        if (trackBr.maxBitrate < minTargetBitrateKbps) {
+            continue
```
adrian-niculescu, line 584: This `continue` skips more than the start hint: the `x-google-max-bitrate` write below is bypassed too, so a track with a sub-300 kbps target loses the codec-level cap it got before this PR. An SVC publish registered at 250 kbps previously produced `x-google-start-bitrate=175;x-google-max-bitrate=250`; now it gets neither. In Rust the equivalent early return is harmless because that munger never writes `x-google-max-bitrate`, but here it does. I'd gate only the `x-google-start-bitrate` append on the threshold (in both the existing-fmtp branch and the `!fmtpFound` one) and keep the max-bitrate write unconditional.

adrian-niculescu:
@xianshijing-lk two housekeeping items for CI: `:livekit-android-test:spotlessKotlinCheck` fails because the modified `SdpMungingTest.kt` needs its copyright year bumped, so run `./gradlew spotlessApply` and push to the branch. And since this changes shipped bitrate and degradation defaults, it needs a changeset file in `.changeset/` (changeset-bot is flagging it too).

livekit-android-sdk/src/main/java/io/livekit/android/room/PeerConnectionTransport.kt
```diff
@@ -469,14 +480,29 @@ fun ensureCodecBitrates(
...
+        }
+
         val fmtps = media.getFmtps()
+        // Use 90% of target bitrate, capped at 1 Mbps for camera to prevent BWE from starting too aggressively
+        // Screen share is not capped since text/UI clarity requires high bitrate from the start
+        // TODO: dynamically adjust start bitrate based on network conditions (e.g., use previous BWE estimate)
+        val calculatedStartBitrate = (trackBr.maxBitrate * startBitrateMultiplier).roundToLong()
+        val startBitrate = if (trackBr.isScreenShare) {
```
adrian-niculescu: There is a structural problem with choosing the cap per track: libwebrtc consumes these fmtp params per connection, not per m-section. In the m144 line this SDK ships (144.7559.05), [`WebRtcVideoSendChannel::ApplyChangedParams`](https://github.com/webrtc-sdk/webrtc/blob/6c1aa903241e69eb2eca64caad16779351bb1ab2/media/engine/webrtc_video_engine.cc#L1301-L1360) reads `x-google-start-bitrate`/`x-google-max-bitrate` via [`GetBitrateConfigForCodec`](https://github.com/webrtc-sdk/webrtc/blob/6c1aa903241e69eb2eca64caad16779351bb1ab2/media/engine/webrtc_media_engine.cc#L160-L183) and pushes the result into the shared Call config, where [`RtpBitrateConfigurator::UpdateWithSdpParameters`](https://github.com/webrtc-sdk/webrtc/blob/6c1aa903241e69eb2eca64caad16779351bb1ab2/call/rtp_bitrate_configurator.cc#L56-L76) replaces the stored config unconditionally, last writer wins. `ApplyChangedParams` even carries a TODO noting codec max bitrate probably should not affect the global call max.

Two consequences:

1. The camera/screenshare split cannot work as written. With both published, the capped camera value and the uncapped screenshare value land in the same SDP, and whichever m-section applies its send parameters last seeds the single BWE. Depending on order, the uncapped screenshare start overrides the camera cap for the whole call, or the camera value clobbers the screenshare hint and the exemption silently does nothing.

2. The max-bitrate write is the sharper edge. A default camera publish now emits `x-google-max-bitrate=2310`, and that value becomes the Call's `max_data_rate`, a ceiling on total send bandwidth across all tracks. If the camera m-section applies last, a concurrent 3 Mbps screenshare is starved under a 2.3 Mbps connection ceiling. Before this PR the conflict required at least one SVC publication (a lone VP9 screenshare's max could already ceiling a concurrent camera); this change brings it to the default VP8 camera plus screenshare case.

Since the knob is per connection, the fix that actually holds is one connection-level value applied consistently to every video m-section, or dropping the global hints entirely; a uniform cap merely bounds the damage, since tracks with different targets still produce different values and the last-writer variance remains. For the max specifically, consider not writing `x-google-max-bitrate` for simulcast at all: the per-encoding `maxBitrateBps` already caps each layer, so the fmtp value only acts as the call-wide ceiling described above. The merged JS change (livekit/client-sdk-js#1987) shares the start-bitrate interaction, though not the max one (the JS munger never writes `x-google-max-bitrate`), so this is probably a cross-SDK decision rather than an Android-only fix.

livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt
```diff
@@ -716,7 +724,12 @@ internal constructor(
                 (track as LocalVideoTrack).codec = finalOptions.videoCodec
 
                 val rtpParameters = transceiver.sender.parameters
+                // Use provided degradation preference, or default based on track source:
+                // - Camera: MAINTAIN_FRAMERATE (smoother video for real-time communication)
+                // - Screen share: MAINTAIN_RESOLUTION (clarity is critical for text/UI)
+                // - Other/unknown: BALANCED
                 rtpParameters.degradationPreference = finalOptions.degradationPreference
+                    ?: getDefaultDegradationPreference(trackSource)
```
adrian-niculescu: Revised, the first version of this comment overstated the default case. The backup transceiver reuses the same RTC track and source, and with no explicit preference libwebrtc's [`GetDegradationPreference`](https://github.com/webrtc-sdk/webrtc/blob/6c1aa903241e69eb2eca64caad16779351bb1ab2/media/engine/webrtc_video_engine.cc#L1930-L1965) resolves `is_screencast` sources to MAINTAIN_RESOLUTION and camera content to MAINTAIN_FRAMERATE, so for camera and screen share the backup sender lands on the same preference implicitly and the new defaults stay consistent. What does not carry over to the backup sender: an explicitly supplied `degradationPreference` (a gap that predates this PR), and for sources outside CAMERA/SCREEN_SHARE the new BALANCED fallback applies only to the primary, while the backup retains libwebrtc's source-derived default (MAINTAIN_FRAMERATE for camera content, MAINTAIN_RESOLUTION for screencast content). If `publishAdditionalCodecForTrack` gets touched anyway, setting the resolved preference on the backup sender would close both.

livekit-android-sdk/src/main/java/io/livekit/android/room/PeerConnectionTransport.kt
```diff
@@ -469,14 +480,29 @@ fun ensureCodecBitrates(
...
+            calculatedStartBitrate
+        } else {
+            minOf(calculatedStartBitrate, maxStartBitrateKbps)
```
adrian-niculescu: This cap regresses the configured start for high-bitrate SVC cameras: previously 0.7x with no ceiling, so a 3 Mbps VP9 camera publish started at 2100 kbps; now it starts at 1000. Above roughly 1.4 Mbps targets the new start is lower than the old one, which can lengthen the ramp on links that can carry the target (on constrained links the lower seed is arguably safer). Screen shares are exempt and go up (2100 to 2700), so it only affects camera SVC. JS ships the same values, so nothing may need to change here, but it is worth stating as a known tradeoff in the PR description, or reconsidering the cap for SVC.

livekit-android-sdk/src/main/java/io/livekit/android/room/PeerConnectionTransport.kt
```diff
@@ -512,6 +538,7 @@ internal fun isSVCCodec(codec: String?): Boolean {
 data class TrackBitrateInfo(
     val codec: String,
     val maxBitrate: Long,
+    val isScreenShare: Boolean = false,
```
adrian-niculescu, line 627: `TrackBitrateInfo.maxBitrate` has always carried kbps, but nothing in the name or type says so, and neighboring fields like RtpParameters' `maxBitrateBps` set a bps expectation. The old test demonstrated the ambiguity: it passed 1000000 into the kbps field and asserted a roughly 1 Gbps fmtp value. Renaming to `maxBitrateKbps` closes the trap. While here: `isScreenShare` could use a KDoc, and the camera/screenshare/other defaults mapping is now written out in five places (both data class comments, the base KDoc, the publish-site comment, and `getDefaultDegradationPreference`); one authoritative copy in the public KDoc would keep them from drifting.

One consideration for both the new `isScreenShare` parameter and the rename: `TrackBitrateInfo` is a public JVM type in the shipped 2.27.0 AAR (`@suppress` only hides it from docs), and changing the primary constructor drops the compiled `(String, long)` constructor and `copy` descriptors. Exposure is low since `PeerConnectionTransport` is internal and the only public entry point taking the type is the `@VisibleForTesting` `ensureCodecBitrates`, but if its ABI is meant to be stable that should be a deliberate call rather than an accidental side effect.

livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt
```diff
@@ -697,14 +697,22 @@ internal constructor(
             track.statsGetter = engine.createStatsGetter(transceiver.sender)
 
             val finalOptions = options
-            // Handle trackBitrates
-            if (encodings.isNotEmpty()) {
-                if (finalOptions is VideoTrackPublishOptions && isSVCCodec(finalOptions.videoCodec) && encodings.firstOrNull()?.maxBitrateBps != null) {
+            // Handle trackBitrates - apply start bitrate for all video codecs to prevent initial blurriness.
+            // - SVC codecs: use first encoding's bitrate (single stream with built-in layers)
+            // - Simulcast: sum all encoding bitrates (independent streams, BWE needs total)
+            if (encodings.isNotEmpty() && finalOptions is VideoTrackPublishOptions) {
```
adrian-niculescu, line 720: A lifecycle gap this expansion widens: `trackBitrates` in PeerConnectionTransport is insert-only, and neither `unpublishTrack` nor `RTCEngine.removeTrack` removes the entry, so registrations outlive their publications for the transport's lifetime. Mostly that is just dead map entries scanned on every offer, but it becomes wrong munging on republish: `cid` is the RTC track id, so republishing the same track hits the same key, and if the new publish computes no encodings (`videoEncoding == null` with `simulcast = false` returns an empty list from `computeVideoEncodings`) this block is skipped and `ensureCodecBitrates` injects the previous publish's start/max values into the new offer. This existed for SVC registrations before, but the all-video path makes it reachable for every codec. An unregister call when the track is removed would close it.

livekit-android-sdk/src/main/java/io/livekit/android/room/PeerConnectionTransport.kt
xianshijing-lk: Yes, the 1,000 kbps cap is intentional for now. We found that setting a high startup bitrate can cause congestion under poor network conditions, which may lead to higher first-frame latency and choppy video during the first ~30 seconds.

Capping the startup bitrate at 1,000 kbps improves that behavior. Under good network conditions, 1,000 kbps is still enough to support 720p video, and the quality should ramp up quickly afterward. Under poor network conditions, for example, when the available bandwidth is below 300 kbps, the bandwidth estimator will converge more quickly and avoid overshooting the available capacity.

I'm planning to follow up by making the startup bitrate more adaptive to the network conditions, using signals such as previous bandwidth estimates, signaling latency, connection quality, or other indicators, instead of relying on a fixed 1,000 kbps cap.

livekit-android-sdk/src/main/java/io/livekit/android/room/PeerConnectionTransport.kt
xianshijing-lk: Thanks for the technical details, they are extremely helpful. I completely missed the webrtc implementation detail that x-google-start-bitrate and x-google-max-bitrate are effectively connection level knobs rather than per-track settings. Really appreciate you digging into the WebRTC internals and pointing this out.


After thinking about it some more, I wonder if the right direction is something like this:
```
x-google-max-bitrate:
    SUM of all active tracks’ max bitrates
    only if LiveKit wants a hard connection-wide ceiling

x-google-start-bitrate:
    MAX of active tracks’ start hints
    only when initializing a new PeerConnection

Track added later:
    recompute max
    leave start unchanged

Track removed:
    recompute max
    leave start unchanged
```

x-google-max-bitrate is effectively a ceiling for the whole PeerConnection, so an individual track's maximum is not sufficient once multiple video tracks are active. For example, if a camera is capped at 2.3 Mbps and a screenshare at 5 Mbps, setting either value alone would incorrectly cap the total connection. If we keep using this parameter, we should maintain the maximum bitrate requirements of all active video tracks, sum them, and apply the same aggregate value to every video m-section.
Note, we might need to apply this x-google-max-bitrate for both simulcast and non-simulcast, since we might have different config combinations of tracks, we will need to update the max-bitrate with new published track otherwise the old obsolete value might constraint the tracks in a wrong way.

x-google-start-bitrate has a different purpose. It seeds the shared bandwidth estimator when a new PeerConnection is initialized; it is not a per-track allocation. For the initial connection, we can use the maximum start hint among the tracks known at that time. I would avoid summing the start hints because that could initialize BWE too aggressively and cause congestion.

Once the PeerConnection is active and BWE has learned something about the network, publishing an additional track should not reset x-google-start-bitrate. The new track should share the existing estimated bandwidth, and WebRTC can probe upward if more capacity is available. We should still recompute the aggregate max bitrate so that the old connection ceiling does not prevent the new track from ramping up.

Similarly, when a track is removed, we should remove its entry from the active-track map and recompute the aggregate maximum. I would leave the start bitrate unchanged because the estimator already has live network information, and lowering or resetting the start hint would not be useful.

What do you think ?

livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt
xianshijing-lk: Done with moving the default degradation preference to a separate PR : https://github.com/livekit/client-sdk-android/pull/991

livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt
xianshijing-lk: You're right that the map is insert-only, and it predates this PR. And I don't think an unregister is needed due to the following understanding.

Both halves of the concern are now gated on hasAppliedVideoStartBitrate. computeConnectionStartBitrate only runs while that latch is false, and ensureCodecBitrates returns before touching the map once it's set, so after the first video offer carries the hint, trackBitrates is never iterated again on that transport , the dead entries aren't scanned on every offer, they aren't scanned at all.

That also bounds the republish case. The latch is set by the first video publish itself, so reaching the stale entry additionally requires that first publish to have failed to set it , setMungedSdp falling back to the unmunged SDP, or a sub-300 kbps target (where the stale entry contributes null anyway). Only then do your conditions apply on top: same track object, and a republish that computes no encodings so no fresh entry overwrites the stale one.

The other half of the original concern is gone outright: the max-bitrate write was removed, so a stale entry can no longer cap the connection. What's left is one start-bitrate hint derived from the same track's previous target, bounded by the 300 kbps floor and the 1 Mbps cap, and self-correcting as the estimator converges.

livekit-android-sdk/src/main/java/io/livekit/android/room/PeerConnectionTransport.kt
xianshijing-lk: This code is outdated, and we don't set x-google-max-bitrate  any more as that is not a correct thing to do.

livekit-android-sdk/src/main/java/io/livekit/android/room/PeerConnectionTransport.kt
xianshijing-lk: Renamed — to targetBitrateKbps rather than maxBitrateKbps, since after dropping the x-google-max-bitrate write the field is only ever a target for the start hint, never a cap.

Confirmed the unit against libwebrtc while I was in there: GetBitrateConfigForCodec does config.start_bitrate_bps = bitrate_kbps * 1000. So the old test asserting x-google-start-bitrate=700000 was claiming 700 Mbps.

isScreenShare now has a KDoc covering why screen shares are exempt from the 1 Mbps cap. The five-places duplication resolved itself — those were all degradation-preference docs, which moved out to #991.

On ABI: agreed it shouldn't be accidental, so I made it deliberate in the other direction — TrackBitrateInfo, TrackBitrateInfoKey and the two-arg ensureCodecBitrates are now internal. They were public only so the separate test module could reach them, and friendPaths (#925) made that unnecessary. Worth noting the visibility change is itself ABI-neutral,  javap shows byte-identical signatures, since Kotlin records internal in @Metadata only and doesn't mangle top-level declarations. The constructor and copy descriptor change from the rename still stands; it's now just unambiguously a change to something that was never API. Called out in the changeset.

xianshijing-lk:
Another ping @adrian-niculescu  @davidliu, could you please help review this PR ?

livekit-android-sdk/src/main/java/io/livekit/android/room/PeerConnectionTransport.kt
adrian-niculescu: Yes, that is the right shape, and it is what landed: one value, the max across the video tracks present in the first video offer, written to every matching video m-section, once per publisher transport, and no max-bitrate write at all. I also checked the once-per-connection latch against reconnect: a full reconnect builds a new transport in `RTCEngine.configure` and the republish re-registers the tracks, so the new estimator is seeded again. Resolving.

livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt
adrian-niculescu: Closed by #991, which applies the resolved preference on the backup transceiver as well. Resolving.

livekit-android-sdk/src/main/java/io/livekit/android/room/PeerConnectionTransport.kt
adrian-niculescu: Correct, and making the types internal is the cleaner call. Resolving.

livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt
adrian-niculescu: Agreed. Checked it against the current head: `computeConnectionStartBitrate` only runs while the latch is false, and `ensureCodecBitrates` returns before touching the map once `connectionStartBitrate` is null, so the stale entries are never read after the first video hint lands. What is left is a start hint bounded by the 300 kbps floor and the 1 Mbps cap. Resolving.

adrian-niculescu (approved):
LGTM.
The latest revision addresses everything from my earlier rounds: a single connection-level start hint written once per publisher transport, no `x-google-max-bitrate` write, the protocol pointer restored, and the degradation defaults split out to #991.

PR #974, opened 2026-06-30
Runtime audio configuration for local audio tracks

davidliu:


hiroshihorie (approved):
LGTM. One small question: is it intended for `LocalAudioTrack.options` to be publicly settable? It looks like callers could update the observable options and advertised features without going through `applyOptions(...)`, so WebRTC would not receive the change. I could be wrong 🙂 Also I assume processing modes are not being exposed in this PR intentionally.

davidliu:
@hiroshihorie ah good catch on that. I'll lock that down.

PR #983, opened 2026-07-26
Fixed ScreenCaptureService staying bound when screen share setup is cancelled

adrian-niculescu:
`ScreenCaptureConnection` records a binding only when `onServiceConnected` arrives, but the binding exists from the `bindService` call onwards. Cancel the coroutine in that window and the `ServiceConnection` stays registered, so `BIND_AUTO_CREATE` keeps `ScreenCaptureService` alive for the lifetime of the context.

`stop()` cannot be the whole answer here. `LocalParticipant.setScreenShareEnabled` awaits the bind inside `setTrackEnabled`, and the track it creates is not published or stored yet, so a cancellation there abandons the track without any `finally` to stop it. Nothing reaches `stop()` at all. A cancelled connect now releases the binding itself once no caller is left waiting on it, and `stop()` unbinds on that same wider condition. Callers stay tracked until `connect()` actually returns, so this covers a cancellation landing after `onServiceConnected` but before the resumed caller runs, which is otherwise the same leak with the binding already connected.

Two smaller problems live in the same window:

- `bindService` registers a connection even when it returns false or throws, which the platform docs call out, so the existing `throw IllegalStateException` path leaked one as well.
- `stop()` while a connect is pending never resumed the queued continuations, so `startForegroundService` suspended forever.

Two more ways a caller could suspend forever came out of tightening the state handling, and are fixed here too. A `stop()` racing a connect could unbind and snapshot an empty waiter set just before the caller enqueued itself, so requesting the bind and registering the waiter now happen under one lock. A failed `bindService` left the state claiming a bind was in flight, which the next caller would then wait on with nothing pending.

Six tests cover the cancellation, rebind, post-connect cancellation, and pending-stop paths; five of them fail on main and pass here. The sixth pins the happy path, so releasing on cancellation cannot tear down a binding a connected caller still holds. `./gradlew test`, spotless, and detekt pass.

Worth flagging separately: cancellation after `createScreencastTrack` can also orphan the new track because `setTrackEnabled` only cleans it up when `publishVideoTrack` returns false; cancellation from `startForegroundService` or `publishVideoTrack` exits before that branch. Cancellation during the service bind happens before capture starts, so it strands the initialized `SurfaceTextureHelper`, `VideoSource`, capturer, and rtcTrack, but not a `MediaProjection`. Cancellation while publication is suspended happens after `startCapture` and can leave the active `MediaProjection` running too. This is outside the `ScreenCaptureConnection` fix. Fully addressing it requires cancellation-safe ownership transfer in `LocalParticipant` and correct `SurfaceTextureHelper` ownership in `LocalScreencastVideoTrack`, and is not attempted here, but happy to take it on if you want it in the same change.

### DIFF
--- a/livekit-android-sdk/src/main/java/io/livekit/android/room/track/screencapture/ScreenCaptureConnection.kt
+++ b/livekit-android-sdk/src/main/java/io/livekit/android/room/track/screencapture/ScreenCaptureConnection.kt
@@ -28,6 +28,7 @@ import kotlinx.coroutines.CancellableContinuation
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.suspendCancellableCoroutine
 import kotlin.coroutines.resume
+import kotlin.coroutines.resumeWithException
 
 /**
  * Handles connecting to a [ScreenCaptureService].
@@ -69,12 +70,23 @@ internal class ScreenCaptureConnection(private val context: Context) {
             }
             connects.forEach { it.resume(Unit) }
         }
+
+        override fun onBindingDied(name: ComponentName) {
+            LKLog.w { "Screen capture service binding died" }
+            failPendingConnects("ScreenCaptureService binding died.")
+        }
+
+        override fun onNullBinding(name: ComponentName) {
+            LKLog.w { "Screen capture service returned a null binding" }
+            failPendingConnects("ScreenCaptureService returned a null binding.")
+        }
     }
 
     /**
      * Binds to [ScreenCaptureService] and suspends until it is connected.
      *
-     * @throws IllegalStateException if the service could not be bound.
+     * @throws IllegalStateException if the service could not be bound, or the binding died or
+     * came back null before the service connected.
      * @throws SecurityException if the caller cannot access the service.
      * @throws CancellationException if [stop] tears the connection down while connecting.
      */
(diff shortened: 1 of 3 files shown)


livekit-android-sdk/src/main/java/io/livekit/android/room/track/screencapture/ScreenCaptureConnection.kt
```diff
@@ -24,76 +24,165 @@ import android.content.Intent
...
+     * True from the start of a bind attempt until its matching unbind attempt. This is a wider
+     * window than [isBound]: the binding is owed an unbind even if the service never connects.
+     */
+    private var isBindRequested = false
     private var service: ScreenCaptureService? = null
-    private val queuedConnects = mutableSetOf<Continuation<Unit>>()
+    private var hasConnectedCaller = false
+    private val queuedConnects = mutableSetOf<CancellableContinuation<Unit>>()
     private val connection: ServiceConnection = object : ServiceConnection {
         override fun onServiceDisconnected(name: ComponentName) {
```
xianshijing-lk, line 52: Is this the only way that binding can fail ?  I wonder if there are other callbacks that we should handle as well

xianshijing-lk (commented):
lgtm in general, one question

livekit-android-sdk/src/main/java/io/livekit/android/room/track/screencapture/ScreenCaptureConnection.kt
adrian-niculescu: The reachable failures are covered: bindService returning false or throwing, both handled in bind(). I went through the rest of the ServiceConnection surface, and for this service the other callbacks cannot fire today:

- onNullBinding fires when the service's onBind returns null. The intent targets the exact ScreenCaptureService class and its onBind unconditionally returns the binder. An app that disables the service via manifest merge lands in the bindService-returns-false path.
- onBindingDied fires when the binding is permanently dead, i.e. the package hosting the service was updated or force-stopped. The service is declared in the SDK manifest without android:process, so it runs in the client's own process and dies with it.
- onServiceDisconnected is handled, and queued waiters intentionally stay queued there: the binding survives a service crash, BIND_AUTO_CREATE recreates the service, and onServiceConnected resumes them.

That said, if onNullBinding or onBindingDied ever became reachable (onBind made conditional, or the service moved out of process), neither is followed by onServiceConnected, so a queued caller would suspend forever, the same bug class this PR fixes. f275e98 handles both as extra caution: release the binding and fail the waiters with IllegalStateException, with a test for each.

xianshijing-lk (approved):
lgtm, thanks

adrian-niculescu:
Thank you for looking at this so swiftly!

davidliu (approved):
LGTM, I think this will need a follow up in LocalParticipant as well, not only to fix up the things mentioned, but cancelling there would also still leak the service binding I think (as far as ScreenCaptureConnection knows, it handled everything fine and it now on the owner to clean up the bind).

PR #986, opened 2026-07-27
Fixed setTrackEnabled orphaning track resources when cancelled before publish completes

### LINKED ISSUE #985: Cancellation during screen share enable orphans track resources after service bind
## Summary
[#983](https://github.com/livekit/client-sdk-android/pull/983) fixes `ScreenCaptureService` staying bound when a coroutine is cancelled *during* `ScreenCaptureConnection.connect()`. A broader follow-up remains: `LocalParticipant.setScreenShareEnabled` / `setTrackEnabled` can still abandon a newly created screencast track on cancellation without calling `stop()` / `dispose()`, leaking WebRTC and capture resources. If cancel happens after a successful bind, the service binding is orphaned too.
## Background
Screen share enable currently does roughly:
```kotlin
val track = createScreencastTrack(...)
track.startForegroundService(...) // suspend: bind + startForeground
track.startCapture()
if (!publishVideoTrack(track, ...)) {
    // cleanup only on publish failure
    track.stopCapture()
    track.stop()
    track.dispose()
}
```
`setTrackEnabled` has no `try` / `finally` around this path. Cleanup runs only when `publishVideoTrack` returns `false`. A `CancellationException` from `startForegroundService` or `publishVideoTrack` unwinds out of the function and drops the local `track` reference without cleanup.
After `startForegroundService` returns, only `LocalScreencastVideoTrack.stop()` → `ScreenCaptureConnection.stop()` releases the binding. Cancel after that point therefore also leaves `ScreenCaptureService` bound for the lifetime of the context.
## What gets orphaned
| Cancellation timing | Service binding | Other resources |
|---|---|---|
| After `createScreencastTrack`, during `connect()` (before it returns) | Released by #983 | Track / `SurfaceTextureHelper` / `VideoSource` / capturer / rtcTrack orphaned |
| After `startForegroundService` returns, before / during `publishVideoTrack` | Orphaned | Same as above; after `startCapture()`, active `MediaProjection` can keep running too |
## Expected behavior
Cancelling `setScreenShareEnabled(true, ...)` (or the surrounding scope) should always tear down any screencast track that was created but not successfully published: stop capture, unbind / stop the screen capture service if it was started, dispose WebRTC objects, and invoke any `onStop` callback as appropriate.
## Suggested direction
- Make ownership transfer cancellation-safe in `LocalParticipant.setTrackEnabled` for `SCREEN_SHARE` (and likely review the camera / mic enable paths for the same pattern): e.g. `try` / `finally` that disposes the unpublished track on cancel or failure.
- Confirm `LocalScreencastVideoTrack` / `SurfaceTextureHelper` ownership so dispose is correct even when publish never completes.
- Add tests that cancel at each suspend point (`startForegroundService`, `publishVideoTrack`) and assert resources are released (and the service is unbound when a bind had succeeded).
## Related
- https://github.com/livekit/client-sdk-android/pull/983 (bind leak during `connect()`; explicitly deferred this follow-up)

adrian-niculescu:
Fixes #985.

`setTrackEnabled` had no cleanup around its suspend points: a cancellation landing in `startForegroundService` or `publishVideoTrack` dropped the freshly created screencast track without stopping or disposing it, and once the bind had succeeded, `ScreenCaptureService` stayed bound for the lifetime of the context (the case #983 could not reach from inside `ScreenCaptureConnection`). Each enable path now ends in a `finally` that tears down a track it could not publish: capture stops, the track is stopped and disposed, the service binding is released, and `ScreenCaptureParams.onStop` is invoked. The camera and microphone paths get the same treatment, with one extra condition. They share the default track with publishers that do not take the source pub lock, such as `republishTracks`, so their cleanup skips a track that is already published: #988 keeps a duplicate publish from stopping a track a concurrent publisher owns, and that has to hold when a cancellation unwinds too, not just when publish returns false.

`publishTrackImpl` negotiates the sender transceiver concurrently with the add track request, so a cancelled or failed request could leave a negotiated transceiver holding its sender and SDP m-section with no publication to unpublish. The sender is now detached whenever no publication is created, which also covers the publish-timeout path. The rollback is deliberately narrow: a publish usually fails because the signal connection died, and that same failure starts the reconnect that renegotiates the publisher, so the rollback runs only while its publisher is connected and idle. Neither `connectionState` nor the reconnect job has moved by the time a dying session fails its pending publishes, so idleness is tracked with a session identity replaced at every boundary; cleanup resuming after a soft reconnect no longer matches and leaves the publisher alone. The check runs on the RTC thread immediately ahead of the mutation it guards. Stopping the transceiver to release its m-section is reserved for video, matching `unpublishTrack`.

Two smaller fixes ride along: `LocalScreencastVideoTrack` never registered its `SurfaceTextureHelper` with the track's `CloseableManager` the way `LocalVideoTrack.createTrack` does, so the capture thread survived `dispose()` even on normal teardown; and a projection stop already dispatched on its callback thread can reach the track after cleanup has disposed it, where `ScreenCapturerAndroid.stopCapture` throws, so `LocalScreencastVideoTrack.stop()` tolerates losing that race instead of crashing the callback thread.

Two behavioral notes: `onStop` on the publish-failure path now runs after teardown rather than before, so a throwing callback cannot skip the cleanup, and it is delivered at most once per enable attempt (the platform ending the projection mid-publish already delivers it; the failure path delivered it again).

Eleven tests cover cancellation at each suspend point, the rollback and each of its limits, single `onStop` delivery, the late projection stop, helper disposal, and the surviving happy path; six fail on main. `./gradlew test`, spotless, and `:livekit-android-sdk:detektRelease` pass (repository-wide detekt is red on pre-existing findings).

### DIFF
--- a/livekit-android-sdk/src/main/java/io/livekit/android/room/RTCEngine.kt
+++ b/livekit-android-sdk/src/main/java/io/livekit/android/room/RTCEngine.kt
@@ -474,6 +474,7 @@ internal constructor(
         lastRoomOptions = null
         participantSid = null
         regionUrlProvider = null
+        endSignalSession()
         abortPendingPublishTracks()
         closeResources(reason)
         connectionState = ConnectionState.DISCONNECTED
@@ -517,10 +518,19 @@ internal constructor(
         client.close(reason = reason)
     }
 
-    private fun abortPendingPublishTracks() {
-        // Ordered ahead of the resumes: each one unwinds a publish into cleanup that reads this,
-        // and on some paths the reconnect is only triggered later, by the socket closing.
+    /**
+     * Marks the current signal session as finished, so publish cleanup that resumes afterwards
+     * leaves the publisher alone.
+     *
+     * Precedes [abortPendingPublishTracks]: each abort unwinds a publish into that cleanup, and
+     * on some paths the reconnect which would otherwise end the session only runs later, when
+     * the socket closes.
+     */
+    private fun endSignalSession() {
         signalSessionState = SignalSessionState(ended = true)
+    }
+
+    private fun abortPendingPublishTracks() {
         synchronized(pendingTrackResolvers) {
             pendingTrackResolvers.values.forEach {
                 it.resumeWithException(TrackException.PublishException("pending track aborted"))
@@ -551,7 +561,7 @@ internal constructor(
         }
         val forceFullReconnect = fullReconnectOnNext
         fullReconnectOnNext = false
-        signalSessionState = SignalSessionState(ended = true)
+        endSignalSession()
         val job = coroutineScope.launch {
             var hasResumedOnce = false
             var hasReconnectedOnce = false
@@ -1219,6 +1229,7 @@ internal constructor(
 
     override fun onClose(reason: String, code: Int) {
         LKLog.i { "received close event: $reason, code: $code" }
+        endSignalSession()
         abortPendingPublishTracks()
         reconnect()
     }
@@ -1238,6 +1249,7 @@ internal constructor(
     override fun onLeave(leave: LeaveRequest) {
         LKLog.d { "leave request received: reason = ${leave.reason.name}" }
 
+        endSignalSession()
         abortPendingPublishTracks()
 
         if (leave.hasRegions()) {
--- a/livekit-android-sdk/src/main/java/io/livekit/android/room/RTCEngine.kt
+++ b/livekit-android-sdk/src/main/java/io/livekit/android/room/RTCEngine.kt
@@ -133,6 +133,7 @@ internal constructor(
         }
         when (newVal) {
             ConnectionState.CONNECTED -> {
+                signalSessionState = SignalSessionState(ended = false)
                 if (oldVal == ConnectionState.DISCONNECTED || oldVal == ConnectionState.CONNECTING) {
                     LKLog.d { "primary ICE connected" }
                     listener?.onEngineConnected()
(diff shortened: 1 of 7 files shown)


adrian-niculescu:
Exercised this on a few devices and refined it over a couple of passes since opening. It's ready for review now.

adrian-niculescu:
https://github.com/livekit/client-sdk-android/pull/988 landed first. Cleaning this up.
L.E. Done.

livekit-android-sdk/src/main/java/io/livekit/android/room/RTCEngine.kt
```diff
@@ -505,6 +518,9 @@ internal constructor(
     }
 
     private fun abortPendingPublishTracks() {
+        // Ordered ahead of the resumes: each one unwinds a publish into cleanup that reads this,
+        // and on some paths the reconnect is only triggered later, by the socket closing.
+        signalSessionState = SignalSessionState(ended = true)
```
davidliu, line 530: nit: this probably be a separate method or just added to the calling sites. it is at the appropriate place in the signal session boundary, but piggybacking this method seems awkward.

davidliu (approved):

livekit-android-sdk/src/main/java/io/livekit/android/room/RTCEngine.kt
adrian-niculescu: Agreed. Pulled it out into `endSignalSession()`, called at the three sites ahead of `abortPendingPublishTracks()` and reused in `reconnect()` for the standalone write. The ordering rule now lives on the new method rather than as a comment inside the abort.

PR #988, opened 2026-07-30
Fix silent mic race between Room.connect audio publish and setMicrophoneEnabled

MaxHeimbrock:
Encountered when trying the android agent starter app. 50% of the time I had to mute and unmute again until the agent and transcription reacted to me talking.

Confirmed the fix in the agent starter app, but had to rebase onto 2.26.0 because TokenRequestOptions gained an agentDeployment parameter in released SDK 2.26.1 (#968). Data-class parameter additions change the constructor signature, so components 2.4.0 — compiled against 2.26.0 — crashes on any released SDK ≥ 2.26.1, including 2.27.0. Latest components + latest SDK is a broken pair for every user right now.

### Background

When connecting with ConnectOptions.audio = true while the app enables the mic as soon as the room reports CONNECTED, both paths published the same default audio track. Room.connect's publish did not take the per-source publish lock, so the losing path's duplicate publish failed and its failure branch stopped the just-published track, leaving the publication unmuted but sending silence until the mic was toggled.

### Changes

- Route Room.connect's audio/video publishes through setMicrophoneEnabled/ setCameraEnabled so they serialize on the source publish lock, where "already published" unmutes instead of failing.
- In setTrackEnabled, don't stop a track whose publish was rejected as a duplicate: it is live and owned by the concurrent publisher.
- Log a warning in publishTrackImpl's previously silent duplicate branch.

### DIFF
--- a/.changeset/swift-frogs-drive.md
+++ b/.changeset/swift-frogs-drive.md
@@ -1,0 +1,5 @@
+---
+"client-sdk-android": patch
+---
+
+Fix silent mic race between Room.connect audio publish and setMicrophoneEnabled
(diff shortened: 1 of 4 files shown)


davidliu (approved):
LGTM!

PR #991, opened 2026-08-03
Change default degradation preference by video source

xianshijing-lk:
## Summary
  - Default video degradation preference by track source when `degradationPreference` is not explicitly set.
  - Camera tracks now default to `MAINTAIN_FRAMERATE`.
  - Screen share tracks now default to `MAINTAIN_RESOLUTION`.
  - Other video sources default to `BALANCED`.
  - Preserve explicit `degradationPreference` overrides.
  - Add tests for camera and screen share defaults.

  ## Testing
  - `git diff --check`
  - Focused Gradle test attempted, but compile failed due existing generated protocol reference issues (`AGENT_ERROR`, `PUBLISH_DATA_TRACK_RESPONSE`, `clientProtocol`, etc.).

### DIFF
--- a/livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt
+++ b/livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt
@@ -734,6 +734,7 @@ internal constructor(
 
                 val rtpParameters = transceiver.sender.parameters
                 rtpParameters.degradationPreference = finalOptions.degradationPreference
+                    ?: getDefaultDegradationPreference(trackSource)
                 transceiver.sender.parameters = rtpParameters
             }
 
@@ -1484,10 +1485,17 @@ abstract class BaseVideoTrackPublishOptions {
     abstract val backupCodec: BackupVideoCodec?
 
     /**
-     * When bandwidth is constrained, this preference indicates which is preferred
-     * between degrading resolution vs. framerate.
+     * Controls how the encoder trades off between resolution and framerate
+     * when bandwidth is constrained.
      *
-     * null value indicates default value (maintain framerate).
+     * - MAINTAIN_FRAMERATE: Prioritizes framerate, reduces resolution if needed
+     * - MAINTAIN_RESOLUTION: Prioritizes resolution, drops frames if needed
+     * - BALANCED: Balances between both
+     *
+     * If not set (null), the SDK uses defaults based on track source:
+     * - Camera: MAINTAIN_FRAMERATE (smoother video for real-time communication)
+     * - Screen share: MAINTAIN_RESOLUTION (clarity is critical for text/UI)
+     * - Other/unknown: BALANCED
      */
     abstract val degradationPreference: RtpParameters.DegradationPreference?
 
@@ -1688,6 +1696,21 @@ internal fun VideoTrackPublishOptions.hasBackupCodec(): Boolean {
 
 private val backupCodecs = listOf(VideoCodec.VP8.codecName, VideoCodec.H264.codecName)
 private fun isBackupCodec(codecName: String) = backupCodecs.contains(codecName)
+
+/**
+ * Returns the appropriate degradation preference for a video track based on its source.
+ *
+ * - Camera: MAINTAIN_FRAMERATE (smoother video for real-time communication)
+ * - Screen share: MAINTAIN_RESOLUTION (clarity is critical for reading text/UI)
+ * - Other/unknown: BALANCED
+ */
+private fun getDefaultDegradationPreference(source: Track.Source): RtpParameters.DegradationPreference {
+    return when (source) {
+        Track.Source.CAMERA -> RtpParameters.DegradationPreference.MAINTAIN_FRAMERATE
+        Track.Source.SCREEN_SHARE -> RtpParameters.DegradationPreference.MAINTAIN_RESOLUTION
+        else -> RtpParameters.DegradationPreference.BALANCED
+    }
+}
 
 /**
  * A handler that processes an RPC request and returns a string
--- a/livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt
+++ b/livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt
@@ -569,11 +569,7 @@ internal constructor(
                 requestConfig = {
                     width = track.dimensions.width
                     height = track.dimensions.height
-                    source = options.source?.toProto() ?: if (track.options.isScreencast) {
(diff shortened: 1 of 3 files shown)


xianshijing-lk (commented):
@adrian-niculescu  @davidliu , I broke down default degradation preference changes from https://github.com/livekit/client-sdk-android/pull/973 to keep the discussion more focused. 

Please review this PR.

davidliu (approved):

livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt
```diff
@@ -1689,6 +1697,21 @@ internal fun VideoTrackPublishOptions.hasBackupCodec(): Boolean {
...
+ *
+ * - Camera: MAINTAIN_FRAMERATE (smoother video for real-time communication)
+ * - Screen share: MAINTAIN_RESOLUTION (clarity is critical for reading text/UI)
+ * - Other/unknown: BALANCED
+ */
+private fun getDefaultDegradationPreference(source: Track.Source): RtpParameters.DegradationPreference {
+    return when (source) {
+        Track.Source.CAMERA -> RtpParameters.DegradationPreference.MAINTAIN_FRAMERATE
+        Track.Source.SCREEN_SHARE -> RtpParameters.DegradationPreference.MAINTAIN_RESOLUTION
+        else -> RtpParameters.DegradationPreference.BALANCED
```
adrian-niculescu, line 1740: A custom feed published with `VideoTrackPublishOptions(source = Track.Source.UNKNOWN)` lands here, and this overwrites what libwebrtc would have derived from the native source: a screencast-backed track goes MAINTAIN_RESOLUTION to BALANCED, a camera-like one MAINTAIN_FRAMERATE to BALANCED. `track.options.isScreencast` already carries that information. BALANCED is also the mode libwebrtc keeps behind the `WebRTC-Video-BalancedDegradation` field trial, with the in-tree note that it ["needs to be tuned first"](https://github.com/webrtc-sdk/webrtc/blob/b1800a61db8320af5c14456c13622d8b85b1ed39/media/engine/webrtc_video_engine.cc#L1954-L1962).

Proposal:

```kotlin
private fun getDefaultDegradationPreference(source: Track.Source): RtpParameters.DegradationPreference? {
    return when (source) {
        Track.Source.CAMERA -> RtpParameters.DegradationPreference.MAINTAIN_FRAMERATE
        Track.Source.SCREEN_SHARE -> RtpParameters.DegradationPreference.MAINTAIN_RESOLUTION
        else -> null
    }
}
```

The KDoc bullet above changes with it. Not a blocker: JS made the same BALANCED choice deliberately, so if this is cross-SDK alignment it stands, but the argument applies there too.

adrian-niculescu (commented):
The backup codec sender never gets the preference.

[`publishAdditionalCodecForTrack`](https://github.com/livekit/client-sdk-android/blob/b3cf15289bb4a089ea2a93f0e3e9056438921403/livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt#L1204-L1255) creates a second transceiver over the same rtcTrack and never touches `sender.parameters`, so that sender keeps resolving implicitly from the native source's `is_screencast`. Consequences:

- Camera and screen share stay media-equivalent across the two encoders as long as `options.source` agrees with `track.options.isScreencast`. A screencast-backed track published with `source = CAMERA` now diverges: primary MAINTAIN_FRAMERATE, backup MAINTAIN_RESOLUTION.
- An application-supplied `degradationPreference` reaches the primary encoder only.
- The new BALANCED fallback reaches the primary encoder only.

The explicit-override half predates this PR and JS has the same gap, but this PR is what makes a resolved preference a value worth carrying, so applying it to the backup sender belongs here.

xianshijing-lk:
> The backup codec sender never gets the preference.
> 
> [`publishAdditionalCodecForTrack`](https://github.com/livekit/client-sdk-android/blob/b3cf15289bb4a089ea2a93f0e3e9056438921403/livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt#L1204-L1255) creates a second transceiver over the same rtcTrack and never touches `sender.parameters`, so that sender keeps resolving implicitly from the native source's `is_screencast`. Consequences:
> 
> * Camera and screen share stay media-equivalent across the two encoders as long as `options.source` agrees with `track.options.isScreencast`. A screencast-backed track published with `source = CAMERA` now diverges: primary MAINTAIN_FRAMERATE, backup MAINTAIN_RESOLUTION.
> * An application-supplied `degradationPreference` reaches the primary encoder only.
> * The new BALANCED fallback reaches the primary encoder only.
> 
> The explicit-override half predates this PR and JS has the same gap, but this PR is what makes a resolved preference a value worth carrying, so applying it to the backup sender belongs here.

Good catch. Confirmed: degradation preference is a sender level property (a top level field on RtpParameters, not per encoding), and publishAdditionalCodecForTrack adds a second transceiver over the same rtc track,  so the backup codec has its own sender with its own parameters, which we were never setting. Worth noting both senders sink from the same VideoSource, and VideoBroadcaster::UpdateWants takes the MIN of max_pixel_count and max_framerate_fps across sinks, so a diverging backup drags its restriction onto the primary too.


Applied the resolved preference to the backup sender, resolving the source the same way the primary publish does rather than reading it back off the publication, so the two can't disagree. Added tests asserting both senders match for the default camera, default screen share, explicit-override, and screencast-backed-published-as-camera cases.

I will port the same fix to other sdks once landing this PR.

livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt
xianshijing-lk: I think BALANCED is fine for the unknown source here.

And balanced mode is set when the source is neither camera nor screen share, which requires an app to explicitly declare something else, source defaults to null and resolves to Camera/ScreenShare  from isScreencast. And it only takes effect when the app hasn't set degradationPreference itself, so anyone who wants a specific behavior (including libwebrtc's implicit derivation) can name it directly.

I ran this locally in both good and constrained network conditions and personally found BALANCED to work pretty well, even without further tuning on the WebRTC side. And for a feed the app has declined to label as motion or detail, a balanced tradeoff seems like the more honest default than committing hard to either axis.

Once this PR is landed, I am going to make follow-up on other SDKs to follow what this PR is doing.

xianshijing-lk:
Hi @adrian-niculescu , thanks for the comments, can you take another look ?

adrian-niculescu (approved):
Read the changes and your comments. LGTM, thanks!

xianshijing-lk:
Thanks @adrian-niculescu and @davidliu

Issue #572, opened 2025-01-02
java.lang.UnsatisfiedLinkError: dlopen failed: library "liblkjingle_peerconnection_so.so" not found


tamimattafi:
**Describe the bug**
The app crashes on some devices (According to Firebase Crashlytics report).
These devices are used by Play Store review team (Installed from `.aab`), we couldn't reproduce this crash using our devices and `.apk` files.

**To Reproduce**
Steps to reproduce the behavior (Version 2.11.0):
1. Create Application class and link it in the Android Manifest
2. Call `Livekit.init(...)` in `onCreate`

**Expected behavior**
The app successfully loads the `.so` library without crashes.

**Stack trace**
```kotlin
Fatal Exception: java.lang.RuntimeException: Unable to create application com.example.app.AndroidApplication: java.util.concurrent.ExecutionException: java.lang.UnsatisfiedLinkError: dlopen failed: library "liblkjingle_peerconnection_so.so" not found
       at android.app.ActivityThread.handleBindApplication(ActivityThread.java:6767)
       at android.app.ActivityThread.access$1500(ActivityThread.java:256)
       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2091)
       at android.os.Handler.dispatchMessage(Handler.java:106)
       at android.os.Looper.loopOnce(Looper.java:201)
       at android.os.Looper.loop(Looper.java:288)
       at android.app.ActivityThread.main(ActivityThread.java:7870)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1009)
       
Caused by java.util.concurrent.ExecutionException: java.lang.UnsatisfiedLinkError: dlopen failed: library "liblkjingle_peerconnection_so.so" not found
       at java.util.concurrent.FutureTask.report(FutureTask.java:123)
       at java.util.concurrent.FutureTask.get(FutureTask.java:193)
       at io.livekit.android.webrtc.peerconnection.RTCThreadUtilsKt.executeBlockingOnRTCThread(RTCThreadUtils.kt:87)
       at io.livekit.android.dagger.RTCModule.libWebrtcInitialization(RTCModule.java:97)
       at io.livekit.android.LiveKit.init(LiveKit.java:72)
       at com.example.app.AndroidApplication.onPreparing(AndroidApplication.kt:42)
[...]

PR #992, opened 2026-08-03
Fix: only latch WebRTC initialization after PeerConnectionFactory.initialize succeeds

shivanshu877:
## Problem

`RTCModule.libWebrtcInitialization` sets `hasInitializedWebrtc = true` **before** calling `PeerConnectionFactory.initialize(...)`. If initialization throws — most commonly when the native library fails to dlopen on a corrupted or partial install (missing arm64 split from sideloaded APKs, app cloners, broken Play delivery; see #572 for the same underlying load failure) — the latch stays set for the lifetime of the process.

The first `LiveKit.create()` then fails with a *catchable* exception (`executeBlockingOnRTCThread` runs the block via `executor.submit(...).get()`, so the `UnsatisfiedLinkError` surfaces wrapped in `ExecutionException`). Apps that catch it and later retry hit the poisoned latch: the second `LiveKit.create()` skips initialization entirely and crashes at the first native call in the Dagger graph:

```
java.lang.UnsatisfiedLinkError: No implementation found for long livekit.org.webrtc.ExternalAudioProcessingFactory.nativeGetDefaultApm()
    at livekit.org.webrtc.ExternalAudioProcessingFactory.nativeGetDefaultApm(Native Method)
    at livekit.org.webrtc.ExternalAudioProcessingFactory.<init>
    at io.livekit.android.webrtc.CustomAudioProcessingFactory.<init>
    at io.livekit.android.dagger.RTCModule.customAudioProcessingFactory
    ...
    at io.livekit.android.LiveKit$Companion.create
```

Because this second failure is a raw `Error` on the caller's thread (not wrapped by a `Future`), typical `catch (Exception)` handling cannot intercept it and the app process dies. We see this pattern in production via Crashlytics.

## Fix

Set `hasInitializedWebrtc = true` only after `PeerConnectionFactory.initialize` returns. A failed initialization stays retryable (webrtc's own `NativeLibrary.libraryLoaded` also remains false when `loadLibrary` throws, so a retry genuinely re-attempts the load), and every failed attempt surfaces as a catchable exception at the `LiveKit.create()` call site instead of a fatal, uncatchable crash on the second call.

## Notes

- Repro: make `System.loadLibrary("lkjingle_peerconnection_so")` fail (e.g. temporarily rename the lib in a test build), call `LiveKit.create()` twice in one process catching `Exception` around each call. Before this change: first call throws `ExecutionException`, second call dies with the `UnsatisfiedLinkError` above. After: both calls throw catchably.
- No behavior change on the success path.

### DIFF
--- a/.changeset/webrtc-init-latch-after-success.md
+++ b/.changeset/webrtc-init-latch-after-success.md
@@ -1,0 +1,5 @@
+---
+"client-sdk-android": patch
+---
+
+Fix: only latch WebRTC initialization after PeerConnectionFactory.initialize succeeds, so a failed native library load stays retryable and surfaces as a catchable exception instead of poisoning the process and crashing on the next LiveKit.create() call.
--- a/livekit-android-sdk/src/main/java/io/livekit/android/dagger/RTCModule.kt
+++ b/livekit-android-sdk/src/main/java/io/livekit/android/dagger/RTCModule.kt
@@ -103,7 +103,6 @@ internal object RTCModule {
         if (!hasInitializedWebrtc) {
             executeBlockingOnRTCThread(LibWebrtcInitializationThreadToken) {
                 if (!hasInitializedWebrtc) {
-                    hasInitializedWebrtc = true
                     PeerConnectionFactory.initialize(
                         PeerConnectionFactory.InitializationOptions
                             .builder(appContext)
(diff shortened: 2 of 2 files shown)


davidliu (approved):
LGTM, can you sign the CLA and add a changeset? Instructions are found here: 
https://github.com/livekit/client-sdk-android/blob/main/CONTRIBUTING.md

shivanshu877:
Done — changeset added and CLA signed. All checks are green; ready for merge. Thanks for the quick review!

PR #993, opened 2026-08-04
Fix custom RTCConfigurations not picking up server-provided ice servers when needed

davidliu:
A [previous variable name change](https://github.com/livekit/client-sdk-android/commit/d0edb9372da83a4857b72d69912f71c2c90e9440) of an outer variable caused the inner apply scope code to refer to the RTCConfiguration's iceServer list (which was empty in this if branch), and not to the server-provided ice servers.

### DIFF
--- a/.changeset/tricky-grapes-compare.md
+++ b/.changeset/tricky-grapes-compare.md
@@ -1,0 +1,5 @@
+---
+"client-sdk-android": patch
+---
+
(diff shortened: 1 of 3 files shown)


MaxHeimbrock (approved):
LGTM

xianshijing-lk (approved):

PR #995, opened 2026-08-04
Sample apps using the Token Source implementations

MaxHeimbrock:
Use Token Sources in our samples: 

`LiteralTokenSource` for the existing path. 

Adding a second tab for connecting via `DevelopmentTokenSource` for:

- sample-app
- sample-app-compose

Screenshot 2026-08-04 at 17 27 23

### DIFF
--- a/sample-app-common/src/main/java/io/livekit/android/sample/CallViewModel.kt
+++ b/sample-app-common/src/main/java/io/livekit/android/sample/CallViewModel.kt
@@ -53,6 +53,8 @@ import io.livekit.android.rpc.RpcError
 import io.livekit.android.sample.model.StressTest
 import io.livekit.android.sample.model.TokenSourceArgs
 import io.livekit.android.sample.service.ForegroundService
+import io.livekit.android.token.ConfigurableTokenSource
+import io.livekit.android.token.FixedTokenSource
 import io.livekit.android.token.TokenRequestOptions
 import io.livekit.android.token.TokenSource
 import io.livekit.android.token.TokenSourceResponse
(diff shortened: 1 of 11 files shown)


sample-app-common/src/main/java/io/livekit/android/sample/CallViewModel.kt
```diff
@@ -149,6 +153,32 @@ class CallViewModel(
...
 
+    /**
+     * Fetches the connection details used to connect to the room.
+     *
+     * The token source is created once and wrapped with [cached], so reconnects reuse
+     * the same credentials until the token expires. Otherwise the development token
+     * server would issue new random values on every fetch when no explicit room
+     * name/participant identity is set, causing reconnects to join a different room.
+     */
+    private val fetchConnectionDetails: suspend () -> Result<TokenSourceResponse> = when (tokenSourceArgs) {
```
davidliu: The documentation and the behavior seems mismatched. It's creating a new token source each time and fetching it.

Can this also be changed to a function that directly returns the `Result<TokenSourceResponse>` rather than creating a lambda? The function declaration should look like this:

```
private suspend fun fetchConnectionDetails(): Result<TokenSourceResponse> 
```

MaxHeimbrock: I pulled out the lambda again and updated the comment to be more specific. The TokenSource is created once and the development token source is using the `.cached` version

davidliu (approved):

davidliu:
Both PRs LGTM!

PR #1004, opened 2026-08-20
feat: Data Tracks

davidliu:
Adds data track support, sharing the Rust UniFFI core (`livekit-uniffi-android`) with the other SDKs.
Most of `room/datatrack/` is a thin wrapper around that core. The Android-specific work is the WebRTC, signaling, and reconnect glue.
**Actual changes**
- `RTCEngine` / `DataTrackFrameSender` — `_data_track` channel, send backpressure, wait-for-open, reconnect
- `SignalClient` — publish / unpublish / subscriber-handle / request-response forwarding
- `Room`, `LocalParticipant`, `RemoteParticipant` — public API, attach/park remote tracks, events
- `IncomingDataTrackManager` / `OutgoingDataTrackManager` — bridge UniFFI managers to the engine
- `DataTrackCryptor` / `E2EEManager` — E2EE via the existing data-packet cryptor
- `DataChannelManager`, `RTCModule`, test DI — channel metering and injectable factories

**Simple wrappers** (FFI type/API mapping, little Android-specific logic)
- `LocalDataTrack`, `RemoteDataTrack`, `DataTrackStream`
- `DataTrackFrame`, `DataTrackInfo`, `DataTrackSid`, `DataTrackSchema`, `DataTrackPublishOptions`, `DataTrackException`
- `DataTrackManagerFactory`, `IncomingDataTrackEvent`

E2E tests: https://github.com/livekit/e2e-android/pull/13


pblazej:
Compared this against the Swift (#975), JS (#1994) and Rust implementations, looking for behavior diffs.

**Before merge**

1. **`DataTrackPublished` is dropped, not deferred** — `Room.kt:905` gates it on `emitWhenConnected`, but `joinImpl` attaches tracks while state is `CONNECTING`/`RECONNECTING`, so it never fires for tracks present at join or after a full reconnect. `DataTrackUnpublished` posts directly (`:821`), so apps can see an Unpublished with no Published. JS deliberately uses plain `emit` for both; Swift notifies unconditionally.
2. **Data-track init can fail `Room.connect`** — `RTCEngine.kt:303` calls `handleSfuJoinResponse` unguarded and builds the UniFFI manager there on every connect, catching only `HandleSignalResponseException`. So an `UnsatisfiedLinkError` fails the connect for apps that never touch data tracks — reachable on API 21–23, where `tools:overrideLibrary` (`AndroidManifest.xml:17`) hides that the AAR won't load and nothing gates on `SDK_INT`. Swift's call is `try?`. `handlePacketReceived` has no try/catch at all and runs on a WebRTC thread.
3. **`waitUntilInboundDataTrackChannelReady()` precedes `resendSubscriptionUpdates()`** — `RTCEngine.kt:781`. If the SFU opens the subscriber channel *in response to* a subscription, the wait can't clear until the call after it runs; at best it adds `MAX_ICE_CONNECT_TIMEOUT_MS` to every full reconnect with remote tracks. Swift, JS and Rust all resubscribe straight after transport connect — suggest deleting the gate. The unit test hand-feeds the channel, so it passes either way.
4. **A dead transport can report `CONNECTED`** — `RTCEngine.kt:362` suppresses `DISCONNECTED` while reconnecting, then the success branch sets `CONNECTED` unconditionally after up to ~40 s of waiting. An ICE failure in between is swallowed and overwritten, and `connectionChangeListener` can't cover it because `reconnect()` early-returns while `reconnectingJob` is live. Not data-track scoped.

**Public surface — cheaper to settle now**

- `DataTrackPublishOptions` (`:22`) allows either field null and never checks compatibility (`JsonSchema` + `Cdr` passes locally, fails remotely). Swift uses two constructors, JS validates against `COMPATIBLE_SCHEMA_ENCODINGS`. A validating factory is additive later, but the permissive constructor is forever. No test covers an invalid pair.
- The `LocalDataTrack` KDoc's "releasing the last reference unpublishes it" is Swift's ARC text — on the JVM the DropGuard waits on the UniFFI Cleaner, so a dropped reference can leave the publication live on the SFU. Nothing is `AutoCloseable` and `close()` only clears the map. I'd fix the doc now and let `AutoCloseable` follow.

**Small**

- `RTCEngine.kt:1079` — the observer wrapper captures the old manager while the gate reads the current field, so a stale `CLOSED` flips an already-open replacement back to `WAITING` and the next publish burns the full timeout. Swift guards with `guard dataChannel === publisherChannel`.
- `IncomingDataTrackManager.kt:62` — the delegate captures its manager and Rust holds it as a JNI global ref, so the cycle spans the FFI boundary and retains the engine and both PCs unless `close()` runs. Swift split out a `weak`-holding shim. Same shape in `OutgoingDataTrackManager`.
- `SignalClient.kt:745` — unguarded `parseFrom` inside the UniFFI forward task throws across the boundary; Swift logs and skips.
- `OutgoingDataTrackManager.kt:78` — `startsWith("Timed out")` to classify the error; reword `RTCEngine.kt:1024/1038` and every timeout silently becomes a disconnect. Two exception types instead.
- `RTCEngine.kt:1580` — subscribe-only clients negotiate a publisher PC via the shared `sendDataTrackSignalRequest`; both references keep subscribe signal-only, and the `hasPublished` guard is redundant (`:814`).
- `DataChannelManager.waitUntilOpen()` has three tests and no production caller; `sendOrQueue`'s per-frame `copyOf()` copies already-fresh packets on the RTC thread.

livekit-android-sdk/src/main/java/io/livekit/android/room/SignalClient.kt
```diff
@@ -112,7 +120,15 @@ constructor(
     /**
      * @see [onReadyForResponses]
      */
-    private val responseFlow = MutableSharedFlow<Pair<WebSocket, LivekitRtc.SignalResponse>>(Int.MAX_VALUE)
+    private val responseFlow = MutableSharedFlow<IncomingSignal>(Int.MAX_VALUE)
```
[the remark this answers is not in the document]
davidliu: This one is tested safe, the Int.MAX_VALUE argument is the replay size, so will effectively playback all the responses that come in during window between JOIN and ready for responses.

davidliu:
@pblazej can I get you to recheck?

xianshijing-lk:
Any plan to land this PR soon ? I got pinged by the Robotic team

livekit-android-sdk/src/main/java/io/livekit/android/room/datatrack/DataTrackSchema.kt
```diff
@@ -0,0 +1,272 @@
...
+        Ros1Msg -> FfiSchemaEncoding.Ros1Msg
+        Ros2Msg -> FfiSchemaEncoding.Ros2Msg
+        Ros2Idl -> FfiSchemaEncoding.Ros2Idl
+        OmgIdl -> FfiSchemaEncoding.OmgIdl
+        JsonSchema -> FfiSchemaEncoding.JsonSchema
+        Other -> FfiSchemaEncoding.Other
+        is Custom -> FfiSchemaEncoding.Custom(identifier)
+    }
+
+    internal fun toProto(): LivekitModels.DataTrackSchemaEncoding {
```
ladvoc, line 129: question: Why does this need to be converted to Protobuf?

livekit-android-sdk/src/main/java/io/livekit/android/room/participant/RemoteDataTrackCollection.kt
```diff
@@ -0,0 +1,140 @@
...
+
+    /**
+     * Adds the track, returning `false` if this exact track is already attached.
+     */
+    fun add(track: RemoteDataTrack): Boolean {
+        val attached = synchronized(lock) {
+            if (tracks.values.any { it === track }) {
+                return@synchronized false
+            }
+            tracks = tracks + (track.name to track)
```
xianshijing-lk, line 51: It uses the track.name as the key ? is this correct ? any chance that two tracks from one participant sharing the same name ? 

or should we use sid as key ?

livekit-android-sdk/src/main/java/io/livekit/android/room/RTCEngine.kt
```diff
@@ -201,6 +198,8 @@ internal constructor(
     private var reliableDataChannelSub: DataChannel? = null
     private var lossyDataChannel: DataChannel? = null
     private var lossyDataChannelSub: DataChannel? = null
+    private var dataTrackDataChannelSub: DataChannel? = null
```
xianshijing-lk, line 201: is this dataTrackDataChannelSub needed ? I don't see how it is used rather than assign and clear, is it intentional ?

livekit-android-sdk/src/main/java/io/livekit/android/room/datatrack/DataTrackManagerFactory.kt
```diff
@@ -0,0 +1,48 @@
...
+import io.livekit.uniffi.RemoteDataTrackManagerInterface
+import uniffi.livekit_datatrack.DecryptionProvider
+import uniffi.livekit_datatrack.EncryptionProvider
+
+/**
+ * Creates UniFFI [io.livekit.uniffi.LocalDataTrackManager] instances.
+ *
+ * @suppress
+ */
+fun interface LocalDataTrackManagerFactory {
```
pblazej, line 31: These are public but their signatures are `io.livekit.uniffi.*` types, and that dep is `implementation` (`build.gradle:121`) — so consumers can't reference them, or the public constructors taking them, without adding it themselves. `@suppress` is Dokka-only, so `internal` is the actual fix; Dagger binds internal types fine within a module.

livekit-android-sdk/src/main/java/io/livekit/android/room/datatrack/OutgoingDataTrackManager.kt
```diff
@@ -0,0 +1,200 @@
...
+            manager.handleSfuRequestResponse(responseBytes)
+        } catch (e: HandleSignalResponseException) {
+            LKLog.w(e) { "Failed to handle RequestResponse for data tracks" }
+        }
+    }
+
+    /**
+     * Republish all tracks after a full reconnect so the SFU recognizes existing publications.
+     */
+    fun republishTracks() {
```
pblazej, line 149: This and `resendSubscriptionUpdates()` run from `coroutineScope.launch` (`RTCEngine.kt:727`), which is `SupervisorJob() + ioDispatcher` with no `CoroutineExceptionHandler` — so anything thrown here hits the thread's uncaught handler instead of failing the reconnect. Does uniffi throw on use-after-destroy if a concurrent `disconnect()` closes the manager mid-call?

pblazej (approved):
LGTM, added 2 non-blockers above

sample-app-common/src/main/java/io/livekit/android/sample/CallViewModel.kt
```diff
@@ -315,6 +292,46 @@ class CallViewModel(
         }
     }
 
+    /**
+     * Subscribes to every data track currently published in the room.
+     *
+     * [RoomEvent.DataTrackPublished] only fires while the room is connected, so tracks that were
+     * already published when we joined — or that were re-attached by a full reconnect — never
+     * announce themselves and have to be picked up from [RemoteParticipant.dataTracks] instead.
```
MaxHeimbrock: This comment contradicts the comment in [RoomEvent.kt](https://github.com/livekit/client-sdk-android/pull/1004/changes#diff-2b8bf94cf25fb2570d679edab31de4307a99460796b260963221f81df2fd8e11R164):
"Fires for every track, including those already published when this participant was first seen and those reattached after a full reconnect."

`RoomEvent.kt` is right about the behaviour, this comment here in the sample is not:


  1. Room drops the data tracks from every remote participant silently, then removes the participants. That is Room.kt:1534.
  2. The engine rejoins. After the join response has created fresh participant objects, it calls the reattach hook at RTCEngine.kt:284. It calls the same
     hook again on every participant update at line 1306.
  3. Room's reattach at Room.kt:834 walks the incoming manager's list of surviving tracks and hands each one to the matching participant at line 822.
  4. The participant's collection accepts the track because the new participant object starts empty. Accepting fires the published callback at
     RemoteDataTrackCollection.kt:55.
  5. That callback posts a participant-level published event at RemoteParticipant.kt:101, and Room turns it into the room-level DataTrackPublished event
     at Room.kt:909.

sample-app-common/src/main/java/io/livekit/android/sample/CallViewModel.kt
```diff
@@ -212,6 +198,10 @@ class CallViewModel(
                 room.events.collect {
                     when (it) {
                         is RoomEvent.FailedToConnect -> mutableError.value = it.error
+                        is RoomEvent.DataTrackPublished -> subscribeToDataTrack(it.track)
+//                        // A full reconnect re-attaches the surviving tracks without republishing
+//                        // events for them, so sweep again rather than waiting on DataTrackPublished.
+//                        is RoomEvent.Reconnected -> subscribeToAvailableDataTracks()
```
MaxHeimbrock: This comment is also stating that full reconnect does not republish events. This should be the case though.

sample-app-common/src/main/java/io/livekit/android/sample/CallViewModel.kt
MaxHeimbrock: Though this seems to still be a valid sweep, because:

"[RoomEvent.DataTrackPublished] also fires for these tracks, but the event collector is started in a sibling coroutine and may not be subscribed yet when connect completes. Sweeping once after connect covers that gap; [subscribedDataTracks] keeps the two paths from subscribing the same track twice."

MaxHeimbrock:
A second draft PR for these changes according to @1egoman's "Commit Reshuffle" strategy to build a story line through commits and some ideas from me for structured PR description for API changes: https://github.com/livekit/client-sdk-android/pull/1019

Please have a look and give feedback to @1egoman and me if this strategy could help review PRs.

livekit-android-sdk/src/main/java/io/livekit/android/room/datatrack/DataTrackSchema.kt
davidliu: They're used when passing the schema over to the server.

livekit-android-sdk/src/main/java/io/livekit/android/room/participant/RemoteDataTrackCollection.kt
davidliu: Name is the stable unique identifier for the track, while sid is the identifier for the current connection of the data track (can change after a reconnect). Name should be the correct key here for lookups, and is what we do on the other SDKs.

livekit-android-sdk/src/main/java/io/livekit/android/room/RTCEngine.kt
davidliu: Hmm, looks like all of the sub channels may need DataChannelManager handling, will handle in a follow up PR.

livekit-android-sdk/src/main/java/io/livekit/android/room/datatrack/DataTrackManagerFactory.kt
davidliu: Keeping this public for direct testing in the E2E testing (can't use internal classes there). Marked it `@VisibleForTesting` to be explicit about why.

livekit-android-sdk/src/main/java/io/livekit/android/room/datatrack/DataTrackStream.kt
```diff
@@ -0,0 +1,130 @@
...
+     *
+     * Concurrent collectors each receive every frame that arrives while they are collecting;
+     * frames are not replayed to a collector that starts late.
+     *
+     * Completes when the stream is unpublished, or when [close] is called.
+     */
+    val flow: Flow<DataTrackFrame> = merge(
+        // onSubscription runs after this collector holds a subscription but before it takes any
+        // value, so a `false` reading here means the terminator is still coming to it.
+        sharedFrames.onSubscription { if (drained.value) emit(null) },
```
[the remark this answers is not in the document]
davidliu: Code is organized such that at least one collector must have collected it for the drained.value to be set, so any future collectors are essentially racing against it anyways.

PR #1007, opened 2026-08-23
feat(audio): negotiate stereo opus on the subscriber answer

### LINKED ISSUE #647: Is it possible to add support for Stereo Channel Count in LocalAudioTrackOptions
**Is your feature request related to a problem? Please describe.**
We are creating a live audio karaoke experience with audio effects, by passing the audio data through some native code through AudioBufferCallback. This native code outputs the audio data in stereo, even when the input is mono. I would rather that the data being input is Stereo to begin with. Since I have to manually convert this data back to mono before sending it to livekit server. 

**Describe the solution you'd like**
If the data coming to onBuffer is Stereo, it would keep make sure there is resampling from stereo to mono. I am hoping we could so something similar to what is happening in JS SDK, where we can set the channel count while creating the local audio track. 

**Describe alternatives you've considered**
For now, i am resampling the audio data back to mono before passing it to livekit. 

**Additional Context**
- https://github.com/livekit/client-sdk-android/blob/68be59fe7032c9374d8d80d1385281987f5f01ca/livekit-android-sdk/src/main/java/io/livekit/android/room/track/LocalAudioTrackOptions.kt#L19
- https://github.com/livekit/client-sdk-js/blob/main/src/room/track/options.ts#L230
- 

Image

kirill-jjj:
## Problem

When a remote participant publishes a stereo audio track, Android subscribers receive it as **mono**. libwebrtc's native Opus decoder downmixes incoming stereo packets to mono unless the local answer negotiates `stereo=1` in the fmtp line — even when the server offer correctly announces the track with `sprop-stereo=1`.

Observed on real hardware (Android 10): the track is published with `stereo=1`, the server relays `sprop-stereo=1` in the subscriber offer — yet the decoded audio plays back mono until `stereo=1` is added to the answer.

## Related issues

- Fixes #647 — that request pointed to `setUseStereoInput(true)`, but that only covers the *publisher* side. The *subscriber* side remained broken.
- Mirrors how client-sdk-js solves this: it collects `remoteStereoMids` from the offer and rewrites the matching fmtp lines when creating the answer (`PCTransport.ts`).

## Changes (receive path only)

- New `StereoSdpMunging.kt`: `ensureStereoOpus()` extracts mids carrying `sprop-stereo=1` from the server offer and adds `stereo=1` to the matching fmtp lines of the answer. Fully fail-safe on parse failures.
- `RTCEngine.onServerOffer` munges the answer before `setLocalDescription`; if setting the munged answer fails, it falls back to the un-munged one (mirrors `PeerConnectionTransport.setMungedSdp`). The fallback lives in a dedicated helper to keep `onServerOffer` complexity in check.
- Changeset (`patch` — no public API changes).

## Scope note (per review)

The publisher-side options (`LocalAudioTrackOptions.stereo`, `TF_STEREO`, `AddTrackRequest.stereo`) were **removed from this PR** and will follow separately, after `useStereoInput`/`useStereoOutput` are exposed by webrtc-sdk/webrtc — so an SDK-level flag can never mismatch what `JavaAudioDeviceModule` is actually configured to do.

## Testing

- 3 unit tests in `livekit-android-test`, using SDP fixtures captured from a real LiveKit session
- Verified on real hardware (Android 10) in a two-participant room: stereo-in confirmed after this change; reproduced mono before it
- Consumer app built (debug + release, R8) and device-tested against the AAR produced by CI
- CI green: spotless, build + unit tests, detekt

_Assisted by AI; verified end-to-end on real hardware._

### DIFF
--- a/.changeset/stereo-answer-munging.md
+++ b/.changeset/stereo-answer-munging.md
@@ -2,4 +2,6 @@
 'livekit-android': patch
 ---
 
-Add stereo=1 to the subscriber answer's Opus fmtp line for media sections where the server offer advertised sprop-stereo=1. Without this, a stereo track published by another participant is decoded as mono on Android.
+Negotiate stereo Opus on the subscriber answer: add `stereo=1` to the fmtp line for media sections where the server offer advertised `sprop-stereo=1`. Without this, a stereo track published by another participant is decoded as mono on Android.
+
+Also adds a `stereo` option to `LocalAudioTrackOptions` (surfaced as the `TF_STEREO` track feature) and `AudioTrackPublishOptions` (sent as `stereo` in the AddTrackRequest), mirroring client-sdk-js `forceStereo`.
--- a/.changeset/stereo-answer-munging.md
+++ b/.changeset/stereo-answer-munging.md
@@ -1,0 +1,5 @@
+---
+'livekit-android': patch
+---
+
+Add stereo=1 to the subscriber answer's Opus fmtp line for media sections where the server offer advertised sprop-stereo=1. Without this, a stereo track published by another participant is decoded as mono on Android.
--- a/.changeset/stereo-answer-munging.md
+++ b/.changeset/stereo-answer-munging.md
@@ -1,0 +1,5 @@
+---
+'livekit-android': patch
+---
+
+Negotiate stereo Opus on the subscriber answer: add `stereo=1` to the fmtp line for media sections where the server offer advertised `sprop-stereo=1`. Without this, a stereo track published by another participant is decoded as mono on Android. Mirrors client-sdk-js behavior.
--- a/livekit-android-sdk/src/main/java/io/livekit/android/room/RTCEngine.kt
+++ b/livekit-android-sdk/src/main/java/io/livekit/android/room/RTCEngine.kt
@@ -54,6 +54,7 @@ import io.livekit.android.webrtc.DataChannelManager
 import io.livekit.android.webrtc.DataPacketBuffer
 import io.livekit.android.webrtc.DataPacketItem
 import io.livekit.android.webrtc.RTCStatsGetter
+import io.livekit.android.webrtc.ensureStereoOpus
 import io.livekit.android.webrtc.copy
 import io.livekit.android.webrtc.isConnected
 import io.livekit.android.webrtc.isDisconnected
@@ -1158,8 +1159,10 @@ internal constructor(
                 return@launch
             }
 
+            val stereoAnswer = answer.ensureStereoOpus(sessionDescription)
+
             run<Unit> {
-                when (val outcome = subscriber?.withPeerConnection { setLocalDescription(answer) }.nullSafe()) {
+                when (val outcome = subscriber?.withPeerConnection { setLocalDescription(stereoAnswer) }.nullSafe()) {
                     is Either.Left -> Unit
                     is Either.Right -> {
                         LKLog.e { "error setting local description for answer: ${outcome.value}" }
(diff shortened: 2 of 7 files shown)


livekit-android-sdk/src/main/java/io/livekit/android/room/track/LocalAudioTrack.kt
```diff
@@ -204,7 +204,12 @@ constructor(
         if (options.autoGainControl) {
             features.add(AudioTrackFeature.TF_AUTO_GAIN_CONTROL)
         }
-        // TODO: Handle getting other info from JavaAudioDeviceModule
+        if (options.stereo) {
+            features.add(AudioTrackFeature.TF_STEREO)
+        }
+        // Note: stereo is sourced from options, not introspected from
+        // JavaAudioDeviceModule; callers that enable stereo input via a module
+        // customizer should set LocalAudioTrackOptions.stereo to match.
```
davidliu: this would need to be surfaced to users, and ideally should be sourced from JavaAudioDeviceModule entirely, to avoid a mismatch in state.

livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt
```diff
@@ -1697,6 +1698,10 @@ data class AudioTrackPublishOptions(
     override val audioBitrate: Int? = null,
     override val dtx: Boolean = true,
     override val red: Boolean = true,
+    /**
+     * Publish the track as stereo, sent as `stereo` in the AddTrackRequest.
+     */
+    val stereo: Boolean = false,
```
davidliu: This should be in BaseAudioTrackPublishOptions, and the stereo option should be added to the end of the list, so that it won't break existing constructors who use positional arguments.

livekit-android-sdk/src/main/java/io/livekit/android/room/RTCEngine.kt
```diff
@@ -1158,11 +1159,24 @@ internal constructor(
...
+                        // subscriber without a local description (mirrors
+                        // PeerConnectionTransport.setMungedSdp).
+                        when (val fallback = subscriber?.withPeerConnection { setLocalDescription(answer) }.nullSafe()) {
+                            is Either.Left -> Unit
+                            is Either.Right -> {
+                                LKLog.e { "error setting local description for answer: ${fallback.value}" }
+                                return@launch
+                            }
+                        }
+                        client.sendAnswer(answer, offerId)
```
davidliu: rather than nesting the fallback here, should return from the run block with an indicator that it should try the fallback in a separate same-level block.

livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt
```diff
@@ -1697,6 +1698,10 @@ data class AudioTrackPublishOptions(
     override val audioBitrate: Int? = null,
     override val dtx: Boolean = true,
     override val red: Boolean = true,
+    /**
+     * Publish the track as stereo, sent as `stereo` in the AddTrackRequest.
```
davidliu: AddTrackRequest is an internal class the consumers don't need to know about.

.changeset/stereo-answer-munging.md
```diff
@@ -0,0 +1,7 @@
+---
+'livekit-android': patch
```
davidliu, line 2: There are API changes in here, so it should be marked as a minor upgrade.

davidliu (changes requested):
This really first needs a PR on https://github.com/webrtc-sdk/webrtc to expose the useStereoInput/useStereoOutput, because a lot of it creates potential state mismatches between what the JavaAudioDeviceModule is actually configured to do and what the user thinks these options will do.

Separately, I think this might be better served as two separate PRs, one for the subscriber side and one on the publisher side.

PR #1008, opened 2026-08-24
fix: treat TokenSource JWTs without exp as expired

SashaMIT:
## Problem

`TokenSourceResponse.hasValidToken` treated a missing `exp` as still valid:

```
val hasExpired = expiresAt != null && now.after(...)
return !isBefore && !hasExpired
```

`TokenSourceCached` then keeps serving that response. A JWT with no `exp` stays cached forever.

`nbf` is already optional (checked only when present). That part is correct. The hole is the suffix-side missing `exp`.

This is the Android cache counterpart of livekit/client-sdk-js#2057. Same class as the server-sdk `TokenVerifier` language-splits (protocol#1706, python-sdks#779, node-sdks#710, server-sdk-ruby#97, server-sdk-kotlin#170). Those require `exp` on verify. This helper decides whether to refetch.

## Fix

Require `exp`. Keep `nbf` optional. Leave the existing one-minute after-expiry tolerance unchanged.

Threat-model: the cached participant token is already on the device. The attacker does not need host or TokenSource authority. A token that omits `exp` never triggers a refetch. The server may still reject the JWT; the client never goes back to the token source.

## Test plan

- `./gradlew :livekit-android-test:testDebugUnitTest --tests io.livekit.android.token.CachingTokenSourceTest` (7/7)
- Revert-tested: restoring `expiresAt != null &&` fails `tokenWithoutExpIsInvalid`
- `./gradlew spotlessApply`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.changeset/token-source-require-exp.md
+++ b/.changeset/token-source-require-exp.md
@@ -1,0 +1,5 @@
+---
+"client-sdk-android": patch
+---
+
+Fix CachingTokenSource treating a JWT with no exp as still valid, so a token that never expires stays cached forever. Require exp; keep nbf optional.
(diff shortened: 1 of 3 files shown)


davidliu (approved):
LGTM, thanks!

SashaMIT:
Thanks for merging this, @davidliu. Glad the no-exp cache path is closed on Android too.

PR #1010, opened 2026-08-25
fix: skip duplicate backup codec publish instead of throwing

### LINKED ISSUE #1000: Exception & crash: VP8 already added
**Describe the bug**
When a track is published with a `backupCodec` (e.g. H265 primary + VP8 backup, `simulcast = true`), a duplicate/late `SubscribedQualityUpdate` from the server can re-enter `LocalVideoTrack.addSimulcastTrack()` for a codec that is already in `simulcastCodecs`, which throws `IllegalStateException: VP8 already added!`. The throw happens inside the SDK's own signal-handling coroutine (Default dispatcher), so it cannot be caught from application code and **crashes the entire app**. In our case (a first-responder drone app), it kills the publisher mid-flight during live streaming.

This appears to be the same crash as #777, which was auto-closed by linear[bot] with no comments.

```
java.lang.IllegalStateException: VP8 already added!
    at io.livekit.android.room.track.LocalVideoTrack.addSimulcastTrack(LocalVideoTrack.kt:~421)
    at io.livekit.android.room.participant.LocalParticipant.publishAdditionalCodecForTrack(LocalParticipant.kt)
    at io.livekit.android.room.participant.LocalParticipant.handleSubscribedQualityUpdate(LocalParticipant.kt)
    at io.livekit.android.room.RTCEngine.onSubscribedQualityUpdate(RTCEngine.kt)
    at io.livekit.android.room.SignalClient (signal response dispatch, SharedFlow collector)
    at kotlinx.coroutines.* (DefaultDispatcher-worker)
```
(Reconstructed from an R8-obfuscated production trace; app on SDK 2.27.0, path verified byte-identical in 2.28.0.)

**To Reproduce**
Steps to reproduce the behavior (intermittent by nature — it is a timing race; hit rate improves with a degraded publisher uplink):

1. Publish a video track from an Android app to a LiveKit Cloud room with:
   ```kotlin
   VideoTrackPublishOptions(
       videoCodec = VideoCodec.H265.codecName,
       backupCodec = BackupVideoCodec(codec = VideoCodec.VP8.codecName),
       simulcast = true,
   )
   ```
   Degrade the publisher's uplink (latency/loss) to stretch the AddTrack round-trip.
2. Have one or more Firefox sessions join and subscribe. Firefox has no H.265 WebRTC decode, so each one makes the server request the VP8 backup publish via `SubscribedQualityUpdate` — the only path that reaches `addSimulcastTrack()`.
3. Briefly interrupt the publisher's network to force a reconnect while the VP8 backup publish is pending / freshly added. Flapping within ~1–2 s of the Firefox join works best; Firefox sessions leaving/rejoining during the degraded period also helps.
4. See error: publisher app crashes with `IllegalStateException: VP8 already added!`.

We reproduced this in-house with 2× Firefox + 1× Chrome subscribers and a publisher-side network flap. No special hardware is required — it should reproduce with the meet example app using the publish options above.

**Expected behavior**
A duplicate publish request for an already-known backup codec should be a no-op, not an app-killing throw. This is exactly what the JS SDK does: `client-sdk-js` has the identical architecture and the identical `!simulcastCodecInfo || !simulcastCodecInfo.sender` detection, but its `addSimulcastTrack` handles the duplicate gracefully ([LocalVideoTrack.ts](https://github.com/livekit/client-sdk-js/blob/main/src/room/track/LocalVideoTrack.ts)):

```typescript
if (this.simulcastCodecs.has(codec)) {
  this.log.error(`${codec} already added, skipping adding simulcast codec`, this.logContext);
  return;
}
```

The Kotlin SDK throws in the same spot instead.

**Screenshots**
N/A (stack trace above).

**Device Info:**
 - Device: DJI RC Plus / DJI RC Plus 2 remote controllers (ARM64); also reproduced on non-DJI hardware
 - OS: Android 10 / Android 11
 - LiveKit SDK version: 2.27.0 (production crashes); crash path verified byte-identical in 2.28.0 and present on current `main`

**Additional context**
Root cause — the "does this codec still need publishing?" check and the "add codec" mutation are not atomic with respect to the async publish round-trip:

1. `LocalVideoTrack.setPublishingCodecs()` treats a backup codec as *new* (needs publishing) when `simulcastCodecs[codec]?.sender == null`.
2. `addSimulcastTrack()` inserts the codec into `simulcastCodecs` **synchronously**, but the entry's `sender` is only assigned later, inside `LocalParticipant.publishAdditionalCodecForTrack()`'s coroutine, **after** the AddTrack server round-trip completes.
3. So there is a window — entry present, `sender` still `null` — during which a second `SubscribedQualityUpdate` for the same codec passes the check in (1), re-enters `addSimulcastTrack()`, and hits `throw IllegalStateException("$codec already added!")`. A reconnect that leaves a stale entry with a null `sender` on a reused `LocalVideoTrack` triggers the same path.

The window is normally milliseconds, which is why the crash is rare — but on a degraded uplink (high RTT/loss, the normal condition for a mobile publisher), the round-trip stretches to seconds and the race becomes realistic. All of our production crash events show connection loss seconds before the crash.

This logic was introduced with SVC codec support in 60ebdd5d (#304, Nov 2023) and is unchanged through 2.28.0 and current `main`.

Proposed fix:
1. Make `addSimulcastTrack()` idempotent — mirror the JS behavior: if the codec is already present, log and return the existing `SimulcastTrackInfo` instead of throwing.
2. Optionally, harden `setPublishingCodecs()`: treat "entry exists with `sender == null`" as *publish in flight* and skip it, rather than returning it as a codec that needs publishing.

Happy to open a PR for (1) if that's welcome.

Related: #777 (same crash, closed without resolution), #304 (introduced the code path)

davibittencourtome:
Fixes #1000

## Problem

When the server sends a `SubscribedQualityUpdate` requesting a backup codec that is already added (or whose publication is still in flight), the SDK crashes the app with:

```
java.lang.IllegalStateException: VP8 already added!
    at io.livekit.android.room.track.LocalVideoTrack.addSimulcastTrack(LocalVideoTrack.kt)
    at io.livekit.android.room.participant.LocalParticipant.handleSubscribedQualityUpdate(LocalParticipant.kt)
    at io.livekit.android.room.RTCEngine.onSubscribedQualityUpdate(RTCEngine.kt)
    at io.livekit.android.room.SignalClient.handleSignalResponseImpl(SignalClient.kt)
```

The exception is thrown on the SDK's internal coroutine dispatcher (`DefaultDispatcher-worker-N`), so there is no way for the application to catch it — it takes down the whole process during a live call.

We hit this in production (self-hosted SFU, publisher on VP9 with the default VP8 backup, `PREFER_REGRESSION`). Repeated identical updates are expected server behavior by design: `MediaTrack.Restart()` (ICE restart / session resume) and `MediaTrack.SetMuted(false)` (unmuting / enabling the camera) both re-emit the current state — `dynacastManager.ForceUpdate()` explicitly skips the "changed" check. A client therefore has to handle a repeated `SubscribedQualityUpdate` idempotently.

## Root cause

There is a window between the two updates in which the codec entry exists but its sender is not yet attached:

1. First update arrives → `setPublishingCodecs` finds no entry for VP8 → returns it as a new codec → `publishAdditionalCodecForTrack` calls `addSimulcastTrack`, which inserts the entry, and then launches a coroutine to create the transceiver. `simulcastTrackInfo.sender` is only assigned inside that coroutine, after the signaling round-trip.
2. Second (identical) update arrives before the sender is attached → `setPublishingCodecs` sees `simulcastInfo?.sender == null` and treats the codec as new again → `addSimulcastTrack` finds the key already present and throws.

The check at `setPublishingCodecs` conflates "never started" with "started, still in flight".

## Fix

Make `addSimulcastTrack` idempotent: if the codec is already present, log a warning and return `null` instead of throwing; `publishAdditionalCodecForTrack` treats `null` as a no-op and returns, so no duplicate transceiver is created and no duplicate `AddTrackRequest` is sent.

This mirrors the JS SDK, where `LocalVideoTrack.addSimulcastTrack` logs and returns `undefined` for an already-added codec and the caller bails out.

Note the dedup is intentionally per track instance (the `simulcastCodecs` map lives on the `LocalVideoTrack`): a track republished after a reconnect is a new track/sid and must still go through backup-codec publication normally. Session-level dedup would break that case.

## Testing

- `addSimulcastTrackForAlreadyAddedCodecIsNoOp` — calling `addSimulcastTrack` twice for the same codec no longer throws; the second call returns `null`.
- `duplicateSubscribedQualityUpdateDoesNotRepublishBackupCodec` — receiving the same `SubscribedQualityUpdate` twice keeps a single backup transceiver (2 transceivers total).
- Full `LocalParticipantMockE2ETest` suite passes.

### DIFF
--- a/livekit-android-test/src/test/java/io/livekit/android/room/participant/LocalParticipantMockE2ETest.kt
+++ b/livekit-android-test/src/test/java/io/livekit/android/room/participant/LocalParticipantMockE2ETest.kt
@@ -341,6 +341,36 @@ class LocalParticipantMockE2ETest : MockE2ETest() {
     }
 
     @Test
+    fun addSimulcastTrackForAlreadyAddedCodecIsNoOp() = runTest {
+        val videoTrack = createLocalTrack()
+
+        val first = videoTrack.addSimulcastTrack(VideoCodec.VP8, emptyList())
+        val second = videoTrack.addSimulcastTrack(VideoCodec.VP8, emptyList())
+
+        assertNotNull(first)
+        assertNull(second)
+    }
+
+    @Test
+    fun duplicateSubscribedQualityUpdateDoesNotRepublishBackupCodec() = runTest {
+        room.videoTrackPublishDefaults = room.videoTrackPublishDefaults.copy(
+            videoCodec = VideoCodec.VP9.codecName,
+            scalabilityMode = "L3T3",
+            backupCodec = BackupVideoCodec(codec = VideoCodec.VP8.codecName),
+        )
+
+        connect()
+        val videoTrack = createLocalTrack()
+        room.localParticipant.publishVideoTrack(videoTrack)
+
+        val trackSid = room.localParticipant.videoTrackPublications.first().first.sid
+        receiveSubscribedQualityUpdate(trackSid)
+        receiveSubscribedQualityUpdate(trackSid)
+
+        assertEquals(2, getPublisherPeerConnection().transceivers.size)
+    }
+
+    @Test
     fun disposeDisposesVideoSource() {
         val source = mock(VideoSource::class.java)
         val videoTrack = createLocalTrack(source = source)
--- a/.changeset/idempotent-add-simulcast-track.md
+++ b/.changeset/idempotent-add-simulcast-track.md
@@ -1,0 +1,5 @@
+---
+"client-sdk-android": patch
+---
+
+Fix crash when a duplicate SubscribedQualityUpdate requests a backup codec that is already added. addSimulcastTrack now skips the duplicate instead of throwing IllegalStateException ("VP8 already added!"), matching the JS SDK behavior.
--- a/livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt
+++ b/livekit-android-sdk/src/main/java/io/livekit/android/room/participant/LocalParticipant.kt
@@ -1273,7 +1273,10 @@ internal constructor(
             return
         }
         val (newOptions, newEncodings) = result
-        val simulcastTrack = track.addSimulcastTrack(codec, newEncodings)
+        // A duplicate SubscribedQualityUpdate can request a codec that is already being
+        // published (or mid-publish, before the sender is attached). Treat it as a no-op
+        // instead of throwing, so a repeated update doesn't crash the session.
+        val simulcastTrack = track.addSimulcastTrack(codec, newEncodings) ?: return
 
         val transceiverInit = RtpTransceiverInit(
             RtpTransceiver.RtpTransceiverDirection.SEND_ONLY,
--- a/livekit-android-sdk/src/main/java/io/livekit/android/room/track/LocalVideoTrack.kt
+++ b/livekit-android-sdk/src/main/java/io/livekit/android/room/track/LocalVideoTrack.kt
@@ -430,9 +430,10 @@ constructor(
         return newCodecs
     }
 
-    internal fun addSimulcastTrack(codec: VideoCodec, encodings: List<RtpParameters.Encoding>): SimulcastTrackInfo {
+    internal fun addSimulcastTrack(codec: VideoCodec, encodings: List<RtpParameters.Encoding>): SimulcastTrackInfo? {
         if (this.simulcastCodecs.containsKey(codec)) {
-            throw IllegalStateException("$codec already added!")
+            LKLog.w { "$codec already added, skipping." }
+            return null
         }
         val simulcastTrackInfo = SimulcastTrackInfo(
             codec = codec.codecName,


jeryini:
Hi @davidliu @MaxHeimbrock @xianshijing-lk . Any chance of getting this fix merged? This is an active issue affecting production. Thanks!

MaxHeimbrock:
I am taking a look

livekit-android-test/src/test/java/io/livekit/android/room/participant/LocalParticipantMockE2ETest.kt
```diff
@@ -340,6 +340,36 @@ class LocalParticipantMockE2ETest : MockE2ETest() {
...
+        connect()
+        val videoTrack = createLocalTrack()
+        room.localParticipant.publishVideoTrack(videoTrack)
+
+        val trackSid = room.localParticipant.videoTrackPublications.first().first.sid
+        receiveSubscribedQualityUpdate(trackSid)
+        receiveSubscribedQualityUpdate(trackSid)
+
+        assertEquals(2, getPublisherPeerConnection().transceivers.size)
+    }
```
MaxHeimbrock, line 371: This test already passes on main today, so it is not testing the fix of your issue. Can you please remove it?

MaxHeimbrock:
Otherwise it looks good and I agree with the proposed changes.

davidliu (approved):
LGTM

livekit-android-test/src/test/java/io/livekit/android/room/participant/LocalParticipantMockE2ETest.kt
davidliu: Looks like there's a pretty specific hard to test window to get the original situation to pass, and the test is a reasonable guard to have overall anyways, so I'm going to keep this one and merge.

PR #1018, opened 2026-09-14
Stop the subscriber entering an ice restart state on a resume

### LINKED ISSUE #1017: Subscriber stops applying remote ICE candidates after an ice restart
`PeerConnectionTransport` sets `restartingIce` to true when it builds an ice restart offer. Only `setRemoteDescription` sets it back to false. After an ordinary resume the server does not send the subscriber a new offer, so the flag stays true for the life of the transport and every candidate after that goes into `pendingCandidates` instead of the peer connection. The subscriber can no longer take any path the server proposes, so the person is still heard and hears nothing.

The same thing was fixed in the web SDK in livekit/client-sdk-js#2054. The conclusion there was that queueing gains nothing, because the server does not send candidates ahead of the offer that introduces them.

Seen on `2.27.0`. The code is unchanged on `2.28.2`.

avosa:
Fixes #1017.

A soft reconnect put the subscriber into an ice restart state, and only `setRemoteDescription` clears that. The server re-offers the subscriber only when the reconnect moved the participant to another node, so after an ordinary resume the flag stayed set for the life of the transport and every later remote candidate was queued instead of applied. The subscriber could no longer take any path the server proposed.

The subscriber no longer enters that state. The server does not send candidates ahead of the offer that introduces them, so queueing them gains nothing, which is the conclusion reached in livekit/client-sdk-js#2054.

The test forces a soft reconnect and asserts a subscriber candidate still reaches the peer connection afterwards.


avosa:
Good catch, that window is real. The hold now goes on in `onServerOffer` before the coroutine is launched, so candidates wait for the offer they belong to and are flushed when its description lands, and nothing holds them across a resume that stays on the same node. Added a test for the window.

avosa:
Fixed, and the read turned up two more paths worth closing.

The hold is now a count rather than a flag, because offer handling runs on a multi threaded scope and two server offers can overlap; a flag let the first one to finish release the hold while the second was still outstanding. Every wait now ends when the attempt to set that description ends, whether it lands, is refused, or answers an offer that has since been superseded, and an ice restart offer that never goes out releases its own.

One deliberate difference from the suggestion: a wait that ends without a description leaves the queued candidates where they are instead of discarding them. A candidate that no longer fits is refused by the connection anyway, while discarding would throw away ones that are still good when the refused description was never going to change the credentials.

PR #1024, opened 2026-09-16
Fixed release publishing failing to build the javadoc jar

adrian-niculescu:
Publishing the SDK fails since data tracks landed: `publishReleasePublicationToMavenRepository`, which the release workflow runs, stops in `:livekit-android-sdk:javaDocReleaseGeneration` with `UnsupportedOperationException: PermittedSubclasses requires ASM9`. CI doesn't catch it because the publish step only runs for snapshot versions.

`withJavadocJar()` makes AGP generate the javadoc with the Dokka version it bundles: 1.4.32 in the AGP 8.7.2 used here, and still in 8.12.1. That Dokka's class reader can't parse the sealed classes in `livekit-uniffi-android`, which is now on the SDK's compile classpath. This replaces the action of AGP's generation task with a sync of the project's own `dokkaJavadoc` (Dokka 1.9.20) output into that task's output directory, so AGP's javadoc jar and the publication's javadoc artifact stay as they are, and output left from an earlier AGP run is removed rather than packaged.

Tested with `./gradlew publishReleasePublicationToMavenLocal -PRELEASE_SIGNING_ENABLED=false -Dorg.gradle.parallel=false`: it fails on main and publishes all four modules with this change, and the SDK's javadoc jar includes the data track classes.


davidliu (approved):
LGTM

PR #1025, opened 2026-09-16
Fixed data track streams dropping frames that arrive just before the end

adrian-niculescu:
Collecting a subscribed data track's `flow` can complete without delivering anything: when the last frames and the end of the stream arrive right as collection starts, the collector sometimes gets none of them. `DataTrackStreamTest.everyFrameSurvivesTheEndOfStream` fails intermittently on main for this reason (1 of 30 runs here).

The drain sets `drained` and then emits its terminator, and `onSubscription` emits a terminator of its own when it reads `drained` as set. A collector that subscribes just before a short drain finishes already has the frames and the terminator buffered, but its `onSubscription` check runs after the flag was set, so its own terminator reaches `takeWhile` first. Pausing 5 ms at the start of that `onSubscription` makes every collection in the test come back empty.

The flag can't tell a collector whether the drain ended before or after it subscribed, but the replay cache can: a collector receives the replayed item atomically with its subscription. The drain now shares wrapped items with `replay = 1`, emitting `Started`, the frames, then `Ended`. A replayed frame predates the collector and is skipped, and a replayed `Ended` completes collection right away, which covers the late collector the flag was there for. The cost is that the replay cache holds the latest frame until the next item replaces it.

With the same 5 ms pause, the test class passes, and without it, it passed 30 runs in a row.


davidliu (approved):
LGTM

